PR #27, opened 2016-04-22
WIP Draft.js

inyono:
**Do not merge me! ;)**

* [x] wrapped in "real" page
* [x] Rich text editing
* [x] LaTeX block equations
* [x] saving (in localStorage) with `CMD+S` resp. `CTRL+S`
* [x] LaTeX inline equations
* [ ] Layout capabilities

Some notes:
* The animation for the LaTeX block equations is not just for show. It is needed to correctly handle click events to equations when one leaves a form above them (otherwise `onMouseUp` is not triggered correctly because the DOM element moved above).
* LaTeX will be rendered on the client. Not sure why the Facebook example had a server. Probably to show splitting the Editor into multiple files without introducing a build system...
* Undo in LaTeX does not work currently, see https://github.com/facebook/draft-js/issues/185


inyono:
superseded by #34

PR #234, opened 2016-11-03
Slate plugins

### LINKED ISSUE #78: core: replace webpack template with create-react-app
To continue our discussion from Monday (I think): If we decide to switch to `create-react-app`, we would lose some syntax stuff (class properties, probably more). Furthermore, css modules aren't supported yet. They work with a little hack, but might break after some updates. On the other hand, we would get a stable config... ;)

### LINKED ISSUE #215: plugins/slate: align text
I should be able to modify the text alignment in blocks (e.g. paragraph, h1) when using slate.

This is a blocker for what I had in mind with #185, so it's a blocker for me and thus high priority

inyono:
Closes #215.
Closes #78.

Included:
* Plugin system for slate
* Refactored npm tasks `build:` and `dev:` for examples using ejected create-react-app config

Things to watch out for:
* [x] There might be problems / missing modules for the new `build:`, `dev:` tasks on Windows, please check @arekkas 
* [x] The live-edit example has currently no production server (dev works though)
* [ ] **Important**. `heroku-postbuild` is probably no longer correct since it probably depended on the building of the examples. Not sure what the process is for this, though.
  * Probably needs `npm run build:home` to build the default example
  * Note that the build now lives in `examples/*/build` instead of `examples/*/public`
* [ ] The editor building process still uses the old webpack config since I'm not sure what you already changed there for npm module building. Will have to review this.


arekkas:
Awesome! The live demo does not work in dev mode as well, unfortunately. But it does not really matter because I removed it anyways with #240 and backed it up here: https://github.com/ory-am/editor-websocket-demo.

Another issue I see is that it is not possible to run multiple web processes on heroku, so we can only run e.g. the home page. I don't think that this is too bad, because it reduces the dependencies and we can easily create a reporsitory with examples. Also the `"start": "cross-env NODE_PATH=. NODE_ENV=production node examples/server.js",` should probably be `npm serve:home` instead.

By the way, I took a very similar approach to the build tasks as you did in #240 :)

Thanks for the changes, I will change the things I mentioned and merge it! :)

Issue #257, opened 2016-11-13
plugins/slate: when slate is focused, undo and redo behave weird


arekkas:

arekkas:
currently the main issue is that when slate is blurred due to an undo action, we're loosing focus in hotkeys as well which breaks ctrl+z. The idea now would be to make hotkeys global and move undoable one level up (basically whole state). Of course, this greatly increases the store size, so we will need our own undoable quite soon...

Also this is an upstream one: https://github.com/ianstormtaylor/slate/pull/421

arekkas:
bäm!

PR #258, opened 2016-11-14
plugins/slate: when slate is focused, undo and redo behave weird #257

arekkas:


arekkas:
one issue persist with selections and using the action toolbar

Issue #252, opened 2016-11-10
plugins: remove class dependency, refactor plugin export


arekkas:
```
export default {
  Component: MyPluginComponent
  name: 'example/content/my-plugin'
  version: '0.0.1'
  icon: <StarIcon />
  text: 'Icon'
}
```

arekkas:
I think it would be better if the hooks where not defined there, but rather in the react component - because then they would have access to the properties natively and we would not have to pass them down. What do you think @inyono ?

inyono:
I don't think that this makes much sense. IMHO, the hooks are on a whole different layer. E.g., they aren't executed when the component is rendered. Furthermore, we don't have access to the mounted component when the hooks are executed. So we would have to instantiate a React component to use the props (and have to pass to it the props, too).

arekkas:
> I don't think that this makes much sense. IMHO, the hooks are on a whole different layer. E.g., they aren't executed when the component is rendered. Furthermore, we don't have access to the mounted component when the hooks are executed. So we would have to instantiate a React component to use the props (and have to pass to it the props, too).

This makes partially sense. We have currently two types of hooks:

* serialize/unserialize
* handleXYZ (handleUndo, handleRemoveKey)

The first class is used outside of component lifecycle, whilst the second isn't. And the second class would need access to the properties of that cell.

However, there is another problem. strg+z is independent of the currently focused cell, you can focus a cell and strg+z something on the other end of the page. That means it will be very difficult to find the cell that the undo operation will take place at - we have to basically inspect the action or something. Nasty.

arekkas:
I think it makes also sense in a scenario where the handleXYZ thing needs access to a DOM reference, or similar. I think this works at the moment too, but has some trickery involved.

arekkas:
Another use case is where ``handleXYZ` need access to a DOM ref inside the plugin. I think there are two possibilities here:

1. Move the handle* things to the component
2. Change the signature of *handle funcs to: `(event: Event, props: PluginProps, component: Component)`

In both cases, we need a registry that maps a unique id to a react component instance. It would allow us to create side effects easily:

```
({id}: action) => {
  const component = lookup(id)
  const handleFocusNextHotKey(.., .., component)
}
```

arekkas:
Additionally I think that plugins should be allowed to export a reducer for the cell state, it would allow us to do something like this:

```
const pluginReducer = (state, action) => {
  if (action.type === 'UNDO') {
    const newState = { ...state }
[...]

PR #259, opened 2016-11-16
WIP Plugin refactor

arekkas:
@inyono this is how we could refactor the plugins. what do you think?


arekkas:
However, this is a rather large refactor because the plugins are used in a lot of places. alternatives are discussed in #252

arekkas:
Also, what's missing is the default values which are defined in LayoutPlugin / ContentPlugin and I'm not sure if there's an easy / sane way to inject those because we might deal with react classes here.

examples/home/src/plugins/video/index.js
```diff
@@ -11,47 +11,51 @@ type Props = {
...
-            <div className="navbar-header">
-              <a className="navbar-brand" href="http://ory.am">ORY</a>
-            </div>
-            <div id="navbar" className="navbar-right">
-              <ul className="nav navbar-nav">
-                <li><a href="https://github.com/ory-am/editor">GitHub</a></li>
-              </ul>
+class Video extends Component {
+  render() {
+    const { children } = this.prpos
```
inyono: typo

arekkas:
Rejected!

PR #262, opened 2016-11-18
core: export editor as AMD & CommonJS module via npm, bower and cdn

arekkas:
@inyono npm run tasks `build:lib` and `build:umd` and `build:umd:min` have been added. Please check if this makes sense!


package.json
```diff
@@ -23,8 +23,11 @@
     "docs": "documentation build ./src/editor/**/*.js --format md --github > API.md",
     "docs:watch": "documentation serve ./src/editor/**/*.js --watch",
     "heroku-postbuild": "npm run build:home",
-    "clean": "rimraf \"public/*\"",
-    "build": "cross-env NODE_PATH=. NODE_ENV=production webpack",
+    "clean": "rimraf \"public/*\" \"dist/*\" \"lib/*\"",
+    "build": "npm-run-all clean build:umd:min build:umd build:lib",
```
inyono: Would change that to

```.json
"prebuild": "npm run clean",
"build": "npm-run-all --parallel build:*"
```

package.json
```diff
@@ -2,11 +2,11 @@
   "name": "ory-editor",
   "version": "0.0.0",
   "description": "The next generation platform for creating well-defined, responsive, digital content - as easy as social media.",
-  "main": "public/bundle.js",
+  "main": "lib/editor/index..js",
   "directories": {
-    "bin": "./public",
+    "bin": "./dist",
```
inyono: What is `bin` doing? I thought this is for CLI tools?

package.json
arekkas: That makes sense, thanks!

package.json
arekkas: Ah, I didn't know that

PR #280, opened 2016-11-21
reducer/editable: adapt tests to new api

### LINKED ISSUE #260: reducer/editable: adapt tests to new api
` [{ id: '000', plugin: 'foo' }]` ~> ` [{ id: '000', plugin: { name: 'foo' } }]`

inyono:
* Updatse dependencies
* Closes #260
* Remove `--coverage` from npm task `test:unit` because it is annoying while developing. Instead, creates new npm task `test:unit:coverage` that will be executed in `test`.


arekkas:
nice!

Issue #265, opened 2016-11-19
plugins/spoiler, reducer/editable: remove defaultPlugin


inyono:
The `defaultPlugin` shouldn't be specified in the Spoiler plugin resp. editable reducer (especially, since we can't customize it).

macrozone:
/close due to inactivity

also there is no spoiler plugin anymore ;-)

PR #304, opened 2017-01-04
0.1.0

arekkas:
@inyono I upgraded some dependencies, most notably `eslint-plugin-react` which deprecated a few rules:

```
The react/no-comment-textnodes rule is deprecated. Please use the react/jsx-no-comment-textnodes rule instead.
The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
The react/wrap-multilines rule is deprecated. Please use the react/jsx-wrap-multilines rule instead.
```

I think those are from the ory eslint dependency. Could you maybe update those?

## todo

* [x] core: have a better createEmptyState #322 
* [x] plugins/spoiler, reducer/editable: remove defaultPlugin #265
* [x] onChange is broken


inyono:
Yep, see https://github.com/ory-am/eslint-config-ory-am-react/issues/3.

inyono:
Sorry, thought this was an issue xD

arekkas:
Still requires inline elements, we probably need to dupe `col-xs-*` for `.inline` and add `width`

arekkas:
@arekkas please write reg test for c05009f

PR #405, opened 2017-06-05
components/Editable: unsubscribe from redux store

### LINKED ISSUE #404: Editable still fires onChange event after <Editable /> component is gone
I have a form that has an `<Editable />` component that I use to submit comments on my page. The render function looks like this: 

```

Class CommentEditor extends Component {

...

render() {
    const document = this.props.document;
    const addValues = this.context.addToAutofilledValues;

    const editor = this.props.editor;
    const setState = this.setState;

    const onChange = function(data) {
      addValues({
        content: data,
      });
      return data;
    };

    return (
      <div className="commentEditor">
        <Editable editor={editor} id={placeholderContent.id} onChange={onChange} />
        <Toolbar editor={editor} />
      </div>
    );
  }

}
```

The `addValues` function calls setState in a higher-level component. When the CommentEditor unmounts, the editable still exists, and makes calls to onChange, even though the component that onChange was modifying doesn't exist anymore. 

Editable should probably remove the editable from the editor on unmount, or there should at least be a convenience function along the lines of `editor.trigger.editables.remove(editableId)` that I could call from my own unmount function.

kacgrzes:
Editable component was missing unsubscribing from redux store on component unmounting. This one should fix #404.

### DIFF
--- a/packages/core/src/components/Editable/index.js
+++ b/packages/core/src/components/Editable/index.js
@@ -17,8 +17,12 @@ class Editable extends Component {
       throw new Error('The state must have an unique id')
     }
 
-    this.props.editor.store.subscribe(this.onChange)
+    this.unsubscribe = this.props.editor.store.subscribe(this.onChange)
     this.previousState = null
+  }
+
+  componentWillUnmount() {
+    this.unsubscribe();
   }
 
   shouldComponentUpdate = shouldPureComponentUpdate


arekkas:
Thank you!

kacgrzes:
@arekkas sir could you please release new version and deploy on npm? :) Thank you!

arekkas:
Sorry, absolutely!

kacgrzes:
Thank you! 🥇

Issue #423, opened 2017-07-16
How to export and restore editor state?


nathanstitt:
I've managed to successfully integrate ORY editor into a fairly large react project and write a few app specific plugins very easily.  I'm very impressed with the work you've done, and think our users will be very happy with the editing experience.  Thanks so much for open-sourcing this!

However, I can't seem to figure out how to save the editor state.  Since ORY's based on slate it is possible to save each content area individually but is there a way to save the entire Editor's state and then restore it?

If you can provide some pointers I'd be happy to make a PR to add to the docs describing how to accomplish this.

arekkas:
I'm glad you were able to gwt it working for your use case! :)

To export it, use the onChange 'onChange={(state) => doSomething(state)}' property, which is a callback, in the Editable component. To restore it, use either the editables config array in 'const editor = new Editor({ editables: [state] })' or, if the editor instance exists already 'editor.triggers.editables.add(state)'.

Im on my mobile so there might be some typos or wrong recollections, but it should point you at the right direction. Check out the example as well, it uses those methods :)

nathanstitt:
Thanks for the awesomely quick reply!  I tested and that works a treat.  I was attempting to muck with `editor.store.getState().editables.present` but that was a object with non-serializable structures.  Figured there had to be a better way 😁 

As promised, I'll make a PR shortly to add the `onChange` to the docs.

Thanks again

nathanstitt:
Circling back to this in case anyone else has the same question.  To set content state for an existing editor it's:
`editorInstance.trigger.editable.add(editorState);`

arekkas:
Hi, yes exactly! I'm also updating some docs right now so this will be documented starting with #450

8ctopotamus:
`editorInstance.trigger.editable.add(editorState);` seems to only work occasionally?

PR #425, opened 2017-07-18
Document saving and restoring editor state

nathanstitt:
As explained in #423, I had trouble figuring this out so thought it'd be nice to have an example.

I tried to stay away from advanced things like class properties and such.  Am happy to rework if you'd prefer any changes or to demonstrate other features.

Thanks for your work on the editor.


arekkas:
Awesome, thank you!

PR #449, opened 2017-09-02
core: implement native drag

arekkas:


zepatrik:
I can't delete the native plugin. Maybe you want to fix this. Works fine apart from that.

arekkas:
Thanks! Resolved. :)

Issue #509, opened 2018-02-03
Feature Request: reducer function for layout plugin


hamavb:
Hello guys,

i noticed a support reducer function of plugin content but not for layout.

since plugin layout also can have a state, it will be usefull to intercept redux actions and alter state basing on some logic, and may spread/sync data to inner plugins content..

PR #510, opened 2018-02-03
plugin: add alternative reducer support for layout plugin

hamavb:
Hello,

I followup a recently added feature request #509 by this PR

My use case is a layout plugin that shares some of its state with other plugins, with some data transformations during synchronization ...
To achieve this, we have created a special HOC that subscribes to ory store, looks for target states based on filter criteria, and maps the results to the current / wrapped plugin. the solution works as expected but it has a performance impact and it is declarative (we prefer it imperative for more control).

By using the option "reducer" of plugin we can achieve the same without the cons of the HOC solution. but unfortunately, the layout plugin does not support it. this PR is about adding "reducer" option support  for the layout plugin .

Please review this PR and consider merging it.

If you have ideas about state sharing between plugins, we are interested in hearing them.

### DIFF
--- a/packages/core/src/reducer/editable/tree.js
+++ b/packages/core/src/reducer/editable/tree.js
@@ -47,7 +47,7 @@ export const cell = (state: Cell, action: Object): Cell =>
           ['content', 'plugin', 'reducer'],
           state
         )
-        const layout = pathOr(identity, ['content', 'layout', 'reducer'], state)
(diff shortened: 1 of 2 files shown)


arekkas:
Thank you for the PR and sorry for the long response time - I'll review it in the evening

arekkas:
Thank you!

PR #518, opened 2018-02-25
Allow images to be links

jtl3d:
This allows one to specify a link location URL which would turn the image into a link. Additionally, a checkbox is provided for opening the link in a new window.

Let me know if there is anything you need me to do to get this in.

### DIFF
--- a/packages/plugins/content/image/src/Component/Form/index.js
+++ b/packages/plugins/content/image/src/Component/Form/index.js
@@ -12,8 +12,15 @@ const handleChange = (onChange: Function) => (e: Event) => {
   if (target instanceof HTMLInputElement) {
     const change = {}
 
-    if (target.type === 'checkbox') {
-      change[target.name] = target.checked ? '_blank' : '';
+    if (target.name === 'target') {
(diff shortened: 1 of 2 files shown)


packages/plugins/content/image/src/Component/Form/index.js
```diff
@@ -2,14 +2,23 @@
...
 import { BottomToolbar } from 'ory-editor-ui'
 
 const handleChange = (onChange: Function) => (e: Event) => {
   const target = e.target
   if (target instanceof HTMLInputElement) {
-    onChange({ src: target.value })
+    const change = {}
+
+    if (target.type === 'checkbox') {
+      change[target.name] = target.checked ? '_blank' : '';
```
arekkas: The issue with `_blank` links is with security, please read more about this [here](https://mathiasbynens.github.io/rel-noopener/) and consider adding the rel attribute!

jtl3d:
Thanks for the tip - added the rel attribute!

arekkas:
Thank you!

PR #527, opened 2018-04-10
feat(doc): add pre-commit githook to regenerate documentation

alexbrbr:
Following discussion on issue https://github.com/ory/editor/issues/524

Here is a precommit hook to run the docs:api on every commit mail. Then the commiter will see the lerna message and the new unstaged files and can decide to add them in a new commit or not (nothing mandatory so it's simpler).

Another option would be to run the doc generation on origin repo after each change on the master branch but I'm not sure how to do this and I lack access to give it a try.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "postcss-cssnext": "^2.10.0",
     "postcss-import": "^9.1.0",
     "postcss-nested": "^2.0.2",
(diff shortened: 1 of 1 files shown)


arekkas:
Thanks! With regards to running this on the CI server, I think it's ok for now to just have it in pre-commit!

PR #529, opened 2018-04-17
Add containerClassName to BottomToolbar, to allow style overrides

HriBB:


arekkas:
Wouldn't it make more sense to have this configurable, similar to `className`?

HriBB:
> That's not as easy, because some of those components make explicit assumptions about, for example, z-index.
I'm not sure I understand this. Haha for sure I don't know enough about the editor yet. Just looked at the code for an hour or so ... 

My idea was, to explicitly require all UI components in user-land, and pass them as config option to `Editor`

```jsx
import Editor, { Editable } from 'ory-editor-core'
import { Trash, DisplayModeToggle, Toolbar, BottomToolbar } from 'ory-editor-material-ui'

const ui = {
  BottomToolbar
}

export const plugins = {
  content: [...],
  layout: [...],
}

const editor = new Editor({
  ui,
  plugins,
  editables: [],
})
```
or as props to `Editable`
```jsx
import Editor, { Editable } from 'ory-editor-core'
import { Trash, DisplayModeToggle, Toolbar, BottomToolbar } from 'ory-editor-material-ui'

export const plugins = {
  content: [...],
  layout: [...],
}

const editor = new Editor({
  plugins,
  editables: [],
})

const AppEditor = ({ content, onChange }) => {
  return (
    <Fragment>
      <Editable
        toolbar={BottomToolbar}
        editor={editor}
        id={content.id}
        onChange={onChange}
      />
      <Trash editor={editor} />
      <DisplayModeToggle editor={editor} />
      <Toolbar editor={editor} />
    </Fragment>
  )
}

export default AppEditor
```

HriBB:
Also

> Wouldn't it make more sense to have this configurable, similar to className?

Where would you pass this configuration? Into `Editor` or `Editable` or ...?

arekkas:
The "core" UI is completely userland right now. The `Editor` class is basically just a container for the redux store.

```
ReactDOM.render((
  <div>
    <Editable
      editor={editor}
      id="something"
    />
    <Trash editor={editor} />
    <DisplayModeToggle editor={editor} />
    <Toolbar editor={editor} />
  </div>
), document.getElementById('controls'))
```

Only plugins have (currently) direct dependencies on the BottomToolbar and similar - which is also why my className suggestion doesn't make sense. I think it would have to be something like:

```
plugins: [slatePlugin({BottomToolbar: MyCustomToolbar}]
```

although this could be a bit difficult to pull off properly, because the API is a bit messy there at the moment and a custom UI component would need to comply with the exact same API contract.

Regarding classnames, something like the following could work:

```
plugins: [slatePlugin({bottomToolbarClassNames: "foo", "bar"}]
```

although it's also not the most beautiful way of solving that...this is tricky.

arekkas:
Closing this because abandoned

PR #530, opened 2018-04-17
Update tutorials.md

simlrh:
Just adding a quick note into the docs, to avoid a head scratching bug I got lost in for a couple of hours!


arekkas:
Thanks!

PR #532, opened 2018-05-01
Slate now exports defaultPlugins

### LINKED ISSUE #482: plugins/slate: export plugins and default plugins
> I used to do this:
```
import slate, { defaultPlugins } from 'ory-editor-plugins-slate'
```
But now defaultPlugins appears to be undefined

> it's now `import { defaultPlugins } from 'ory-editor-plugins-slate/lib/hooks'`

Instead, the default plugins should be exported directly by ory-editor-plugins-slate

davidfurlong:
Fixes #482


arekkas:
Thanks!

PR #533, opened 2018-05-03
Fix previous state to prevent unnecessary `onChange`s

sidati:
Hi @arekkas, As you know jJavaScript is not "natively" smart when it comes to Objects comparison : 

& since ORY has `lodash` dependency i thought using `lodash.isEqual` is good option beside and i make use of `this.previousState` var to stop unnecessary `onChange` events.

### DIFF
--- a/packages/core/src/components/Editable/index.js
+++ b/packages/core/src/components/Editable/index.js
@@ -5,12 +5,12 @@
  * it under the terms of the GNU Lesser General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *  
+ *
  * ORY Editor is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
- *  
+ *
  * You should have received a copy of the GNU Lesser General Public License
  * along with ORY Editor.  If not, see <http://www.gnu.org/licenses/>.
  *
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Editable/index.js
```diff
@@ -47,6 +48,7 @@ type PropTypes = {
 class Editable extends Component {
   constructor(props: PropTypes) {
     super(props)
+    this.previousState = !props.id ? null : editable(props.editor.store.getState(), { id: props.id })
```
arekkas: I think this can lead to a race condition where the component is being mounted but the editor state is not yet available. That's the reason why previousState was initialized with null

packages/core/src/components/Editable/index.js
```diff
@@ -79,10 +81,12 @@ class Editable extends Component {
     const state: any = editable(this.props.editor.store.getState(), {
       id: this.props.id
     })
-    if (state === this.previousState || !state) {
+
+    if (isEqual(state, this.previousState) || !state) {
```
arekkas: This may have some serious performance issues, as this compares every key in the tree with one another. In large documents, this will cause a ton of comparisons. If you type fast, it will cause lags on slow systems. A better idea would be to have immutable objects and do a simple comparison instead.

sidati: What about using JSON.stringify ? Ill try to do some tests

packages/core/src/components/Editable/index.js
sidati: Not sure if i get it 😅, how this can be happened ?

packages/core/src/components/Editable/index.js
arekkas: It will be even worse. What we really should use is [immutable.js](https://facebook.github.io/immutable-js/). You can however use this fix in your own code to prevent the issue you're experiencing. I don't think it makes sense to merge this patch here though.

packages/core/src/components/Editable/index.js
arekkas: Only because the editable is being rendered does not mean that the store has the editable already - it might be loaded asynchronously from a backend or the js engine might decide to execute this code before updating the internal store.

PR #534, opened 2018-05-23
Add babel-preset-react-app as project dependency

stephanebelazouz:
It seems we need babel-preset-react-app to build project but it is not into package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "devDependencies": {
     "babel-cli": "^6.26.0",
     "babel-eslint": "7.2.3",
+    "babel-preset-react-app": "^2.2.0",
     "chokidar-cli": "^1.2.0",
     "coveralls": "^2.13.1",
     "cross-env": "^5.0.0",


arekkas:
Can you please describe the issue that you're facing? I don't think it's required, otherwise test, deployments, packaging, literally anything would have failed :)

stephanebelazouz:
Hi arekkas, thank you for your answer.

Since I use npm instead of yarn this is an example of error output:
`"Can not find the preset" reac-app "relative to the directory" *****/node_modules/ory-editor "`

In the root directory you can find in the yarn.lock: 1198: babel-preset-react-app@^2.2.0
In the subproject example, you can find in the package.json: 29 "babel-preset-react-app": "^ 2.2.0"

But it was not added to the main package.json.

I let you appreciate the need to add it or not :)

And thank you for this beautiful project and for all the work already done with care.

arekkas:
Which command are you using that causes the error? Would it be possible for you to show a reproducible case, starting with `git clone`, then probably `npm i`, ...

Maybe it's an issue with the workflow you're using that can be resolved without code changes!

stephanebelazouz:
Visibly i see in .babelrc that babel uses explicitly react-app preset which is not declared as dev dependencies in package.json.

This incoherence causes an error in install procces (npm i) if no *-lock.json file present.

If this's not yet clear for you i may reproduce it in clone repo later this weekend.

arekkas:
Ah yeah, so the preset is in the yarn lockfile but not in the npm one. This makes sense, just have to figure out why travis is failing

stephanebelazouz:
It's seems i've added a higher version of babel-preset-react-app ^3.x previously ^2.x. This is something could be investigate but this solve test failure due no transpiled file during build.

arekkas:
Perfect, thanks!

PR #539, opened 2018-06-18
Add ory-plugins-content-slate-container class to outer slate div

davidfurlong:
Currently it is hard to define styles for this container due to no class name. Rather than putting in an issue, thought I would add it. This might have been intentional behaviour / ory user defined behaviour, although I couldn't find any docs for this.

Also I tried to remain consistent with the naming convention I saw, although I'm not 100% certain 'container' is the preferred word in this repo


arekkas:
Thank you! I think `ory-plugins-content-slate` will do just fine! Could you please update the static render tests which currently fail?

arekkas:
Thanks!

Issue #553, opened 2018-08-11
Upgrade slate to latest version


PeterKottas:
Hi, this is an issue I am reporting while working on this in case anybody want to contribute anything. I've noticed some warnings that has to do with slate being a rly old version. I wanted to upgrade it (which I though would be straight forward) but it opened whole can of worms as they've changed almost the entire api. After a few hours on this, it seems that the best solution (that doesn't include breaking changes) would be to create brand new slate-next content plugin that would use latest slate. Is this something that would make sense to you? In the meantime I am trying my best to just upgrade and keep it backward compatible but I am struggling. 
In particular, the plugin system changed, and I am not exactly sure how it worked before. The serialize and deserialize on every plugin doesn't make much sense to me at the moment.

arekkas:
Yeah good point - keeping up with slate's changes has been challenging and a lot of breaking changes were introduced. The goal should be to keep the document model somewhat similar. A big breaking change will be upgrading the undo/redo. As long as the document model from the current slate version is compatible (or convertible) to the new one, we should avoid creating slate-next. I can probably find some time to look into this in next week or the week after. Not sure if @inyono has upgraded the slate version and has insights to share?

arekkas:
Oh right, the plugins will probably break. I think that is a breaking change we can live with. What should not be breaking is existing documents!

PR #554, opened 2018-08-12
Upgrade to latest slate

PeterKottas:
**This work is not ready yet.**
I am just opening PR for convenience so that you can see my changes and have a chance to react. 
This of course has to do with #553 
Right now I went with creating slate-next. The idea is to either at the end delete old slate and rename this back to slate. That is of course if we manage to make it work with old documents.


PeterKottas:
Complete list of changes in the underlying serialized data 

- **kind** is renamed to **object**
- **ranges** is renamed to **leaves**
- whole content is wrapped in additional **document** node

I think we could make it backwards compatible by hacking around with https://github.com/PeterKottas/editor/blob/master/packages/plugins/content/slate-next/src/hooks.js#L130
Namely we could update plugin version, check in that place if it's the old version, and if so, just perform these actions on the serialized document. Not that importFromHtml way of storing data doesn't suffer from this problem.

arekkas:
Do you think it would be possible to migrate those changes to the `slate` plugin itself and offer a way for migrations? We do have a way to check for plugin versions iirc for this specific use case.

PeterKottas:
Yeah absolutely, it's already implemented in #557  ;)

PeterKottas:
Just need to wait for that branch to merge, then merge it into this and make the slate-next -> slate renaming, it should then work perfectly with old data. I've also added some unit tests for migrations

arekkas:
Ok I see, merging unstable/dev work into master is not a good practice so I created a new branch https://github.com/ory/editor/tree/slate-next where both PRs can be merged into (please change base of both) - then review it and then finally merge into master. Sound good?

PeterKottas:
Will I be able to commit directly to it?

PeterKottas:
If not I can merge my 2 together and PR against that new branch. I just wanted to keep those separate as the migrations are fairly self contained thing

arekkas:
You don't have write access to this repository at the moment but if you keep up this amazing work you should! If you rebase both PRs against that branch I can merge them right now and you can create another PR against that new branch with the merged changes to make your latest changes. Alternatively you can merge your two branches on your fork and create a new PR. Whatever you like

arekkas:
I can also merge the two PRs on the new branch and you fork it and work on that base

PeterKottas:
Cheers man, yeah I am in a rush to add CMS to our system and as we love ory editor, we have decided to give some love back ;) For now, I'll be happy to merge my 2 together, that will be straight forward. It however makes sense to me to keep the PR for migrations alive, just for your convenience. That way, you can easily see (and discuss) in separation how I implemented migrations. Sounds good?

arekkas:
Sound good! I won't be able to review until end of the week though. The PRs are rather large and it's hard to see what actually changed in this one as it's a completely new code base.

PeterKottas:
Yeah I know and understand. Don't worry about it. Just as soon as possible will do ;) Thanks for letting me know about time schedule

PeterKottas:
Ok I think it's ready for a review. I was checking out the code dif ... and tbh, it's not half as bad as I expected. Spend ages working on it though :D

packages/core/src/service/plugin/index.js
```diff
@@ -162,6 +176,23 @@ export default class PluginService {
     ...this.plugins.layout.map(({ name }: Plugin) => name)
   ]
 
+  migratePluginState = (state: any, plugin: Plugin, dataVersion: string): ?Object => {
+    if (!plugin || !dataVersion || !semver.valid(dataVersion)) {
+      return state
+    }
+    const usedMigrations = plugin.migrations ? plugin.migrations.filter(m => semver.gt(m.version, dataVersion)) : []
+    usedMigrations.sort((a, b) => semver.lt(a.version, b.version) ? 1 : -1)
+    try {
```
arekkas: A particular reason for try/catch here?

packages/core/src/service/plugin/index.js
```diff
@@ -186,18 +217,54 @@ export default class PluginService {
...
+        }
       }
     }
 
     if (layoutName) {
-      const plugin = this.findLayoutPlugin(layoutName, layoutVersion)
-      newState.layout = {
-        plugin,
-        state: plugin.unserialize(layoutState)
+      const found = this.findLayoutPlugin(layoutName, layoutVersion)
```
arekkas: This code block and the previous are very similar. Is it possible to merge those two?

packages/core/src/service/plugin/index.test.js
```diff
@@ -24,9 +24,24 @@ import React from 'react'
 import unexpected from 'unexpected'
 
 import PluginService from './index'
+import { Migration } from './classes';
+
+const FOO = 'foo'
+const WRONG_CONTENT_VERSION = '0.0.0'
```
arekkas: What's the difference between wrong, extra, and valid?

packages/core/src/service/plugin/index.js
```diff
@@ -141,17 +141,31 @@ export default class PluginService {
...
    */
-  findContentPlugin = (name: string, version: string): ContentPlugin => {
+  findContentPlugin = (name: string, version: string): { plugin: ContentPlugin, pluginWrongVersion?: ContentPlugin } => {
     const plugin = this.plugins.content.find(find(name, version))
-    return plugin || new ContentPlugin(contentMissing({ name, version }))
+    let pluginWrongVersion = undefined
+    if (!plugin) {
+      pluginWrongVersion = this.plugins.content.find(find(name, '*'))
+    }
+    return {
```
arekkas: So the API return value changed here, right? What's the the difference between `plugin` and `pluginWrongVersion`?

packages/plugins/content/slate/src/Component/index.js
```diff
@@ -91,32 +88,30 @@ class Slate extends Component {
...
       ToolbarButtons,
       focus
     } = this.props
-    const isOpened = editorState.isExpanded && editorState.isFocused
+    const isOpened = editorState.selection.isExpanded && editorState.isFocused
 
     return (
       <div>
-        <Portal isOpened={isOpened} onOpen={this.handleOpen}>
+        <Portal onOpen={this.handleOpen}>
```
arekkas: react-portal v4 still has the `isOpened` property (it's now called `isOpen`). Any particular reason why you removed this and put it as a class down below?

packages/plugins/content/slate/src/Component/index.js
```diff
@@ -136,11 +131,11 @@ class Slate extends Component {
           onKeyDown={onKeyDown}
           readOnly={Boolean(readOnly)}
           className="ory-plugins-content-slate-container"
-          onBlur={onBlur}
-          schema={schema}
-          state={editorState}
+          onBlur={this.onBlur}
```
arekkas: Any particular reason for moving `onBlur` to `this`? It doesn't seem like we need the classes context for the blur function here?

arekkas:
OMG! Just checked out the demo and this is really amazing, you made it all work! :) Only one thing I noticed while checking it out is that the plugin is called "New text" but that's literally just one fix in the code:

image

I will still go through your changes line-by-line with a hawkish eye as I usually do :)

packages/core/src/service/plugin/index.js
PeterKottas: The migrations can be coming from 3RD party plugin plugin yet we are executing them in the core package. Might be a bit paranoid but as started as backend dev, I have this thing where I wrap 3rd party code in try catch. People might be installing lot of plugins from weird sources and it potentially makes sense for it not to throw down the whole editor. Maybe we could introduce ErrorBoundaries as well. Or maybe I am just ready for my tinfoil hat :D

packages/core/src/service/plugin/index.test.js
PeterKottas: Wrong is one less (older version), valid is the one we actually use, extra is one more (newer version). The whole change should only work when in data the version is let's say 0.0.1 and your installed is 0.0.2. If your installed is 0.0.0 it will not work and fallback to plugin missing.

packages/core/src/service/plugin/index.js
PeterKottas: plugin is the exact version, pluginWrongVersion is higher version of the plugin. So let's say you installed 0.0.2 and the data asks for 0.0.1. The plugin will be populated with the MissingPlugin fallback while the pluginWrongVersion will be populated with 0.0.2 version. You can then devide what to do. Yeah the api changed here but I think this is private api, right? at least I didn't find a place where it was exposed.

packages/plugins/content/slate/src/Component/index.js
PeterKottas: I didn't like (visually) how when you click somewhere it fades in nicely. But when you click out, it just disappears instantly. This was it will do nice animation both ways. That's the only reason

packages/plugins/content/slate/src/Component/index.js
PeterKottas: Just best practice for me (obviously everybody has different sort of rules). In the future, the onBlur might need to use the class even-though now it doesn't. Also it tend to be more straight forward in unit testing. But this obviously is mostly matter of preference. No real harm or benefit comes from this.

PeterKottas:
Cheers mate. Yeah it was crazy at times. I though it will be 5 minutes fix but it ended up being a few days :D You're right about the New text, I'll be fixing this right away.

packages/core/src/service/plugin/index.test.js
arekkas: The example you gave is just for this specific test case, right? Otherwise we should probably use semver. So `^1.0.0` matches `1.*.*`,  `~1.0.0` matches `1.0.*`  and so on. This would also imply that a plugin has a specific version as well as a "compatibleWithVersion" which would be a semver constraint. This should be refactored.

Also, let's call the constants here `OLDER_VERSION`, `MATCHING_VERSION`, and `NEWER_VERSION`. This makes it much easier to read.

packages/core/src/service/plugin/index.js
arekkas: The idea is ok, but the execution needs improvement. We should definitely catch this in an error boundary. I think just throwing out a `console.error` and returning `undefined` (which is a very difficult type do handle) is not the correct approach here. Let's not catch the error and try to (in a different PR) introduce e.g. ErrorBoundary.

packages/plugins/content/slate/src/Component/index.js
arekkas: Personal preference is absolute legitimate. The thing is that large PRs are incredibly difficult to review. Non-functional code changes that have to do with personal preference make it tougher to review as nothing functional has changed, just the style of the code. It's also important that the same code-style is kept throughout the project. This is incredibly hard to achieve but it's a bit easier if contributors try to avoid "personal preference" code-changes here and there.

Rule of thumb is, if a method doesn't require the classes context, move it to a constant. It's more likely that you'll remove the need for the class (specifically in React) and end up with a stateless component than potentially needing the context in some function later. My rule is always: "If in doubt, leave it out".

packages/core/src/service/plugin/index.js
arekkas: Not sure if I understand the API correctly. I'd suggest to go with the idea raised in comment https://github.com/ory/editor/pull/554#discussion_r210720052 . It would remove the need for fallbacks and give plugin authors the ability to define version ranges themselves. If the version range for the installed plugin does not match the version from the content model, we show a plugin not found error, or a plugin does not match content version error.

packages/core/src/service/plugin/index.test.js
PeterKottas: Yeah sure, the constants are only for test cases but I can change those. The changes to semver however, I am not sure about those. Tempted to leave those maybe for a different PR cause it feels like it can open a can worms.

packages/core/src/service/plugin/index.js
PeterKottas: I'll try

packages/core/src/service/plugin/index.js
PeterKottas: Ok

packages/core/src/service/plugin/index.test.js
arekkas: I don't think it will be incredibly difficult, but it's ok to reserve it for another PR. Then my suggestion would be to just do a strict version match `0.0.1 === 0.0.1`, `0.0.1 !== 0.0.2` and remove the custom interpretation of version ranges. If the content version is smaller than the plugin version `0.0.1 <= 0.0.2`, we run it through migrations and migrations is expected to handle it (or throw an exception if not able to).

packages/plugins/content/slate/src/hooks.js
```diff
@@ -61,7 +63,18 @@ export const defaultPlugins = [
 export const lineBreakSerializer = {
   deserialize(el: any) {
     if (el.tagName.toLowerCase() === 'br') {
-      return { kind: 'text', text: '\n' }
+      return { object: 'text', text: '\n' }
```
arekkas: Where is this serializer taken from?

packages/plugins/content/slate/src/hooks.js
```diff
@@ -76,23 +89,29 @@ export const html = new Html({
...
-              text: ''
-            }
-          ]
-        }
-      ]
+      document: {
+        nodes: [
+          {
+            object: "block",
+            type: "paragraph",
```
arekkas: Is it intentional that this is `"paragraph"` and not `P`?

packages/plugins/content/slate/src/migrations/v002.js
```diff
@@ -0,0 +1,31 @@
+import { Migration } from 'ory-editor-core/lib/service/plugin/classes';
+import rename from 'deep-rename-keys';
+const migration = new Migration({
+    version: '0.0.2',
```
arekkas: I think it would be better to define the version ranges here:

```
new Migration({
  toVersion: '0.0.2',
  fromVersionRange: '~0.0.1' // just an example,
  migrate: state => state
})
```

packages/plugins/content/slate/src/plugins/lists.js
```diff
@@ -24,7 +24,6 @@
 import React from 'react'
 import ListIcon from '@material-ui/icons/List'
 import OrderedListIcon from '@material-ui/icons/FormatListNumbered'
-import createListPlugin from 'slate-edit-list'
```
arekkas: Is `slate-edit-list` not compatible with the latest slate or what is the reason for removing it?

packages/plugins/content/slate/src/plugins/paragraph/node.js
```diff
@@ -21,36 +21,13 @@
  */
 
 import React, { Component } from 'react'
-import { Placeholder } from 'slate'
-import { placeholder } from '../../const.js'
-import shallowEqual from 'fbjs/lib/shallowEqual'
 
 class Paragraph extends Component {
-  shouldComponentUpdate(nextProps) {
```
arekkas: A particular reason for removing `shouldComponentUpdate`?

packages/plugins/content/slate/src/plugins/paragraph/node.js
```diff
@@ -21,36 +21,13 @@
...
-  }
-
   render() {
-    const { node, state, children, attributes } = this.props
+    const { children, attributes } = this.props
     const align = this.props.node.data.get('align')
     return (
       <p {...attributes} style={{ textAlign: align }}>
-        <Placeholder
-          className="ory-plugins-content-slate-paragraph-placeholder"
```
arekkas: Particular reason for not keeping the classname?

packages/plugins/content/slate/src/hooks.js
PeterKottas: From docs, just needed to modify it to support the actual text nodes

packages/plugins/content/slate/src/hooks.js
PeterKottas: nope that's an error on my side, fixed

packages/plugins/content/slate/src/migrations/v002.js
PeterKottas: This is actually not how it works right now, it's little bit complicated like always with migrations. I don't even believe there's right way to do this.
Basically right now, you as an author should provide all migrations necessary to migrate from older to newer. 
These are sorted by version. That means that if I define migrations for 
0.0.1 
0.0.3
0.0.6
My plugin is version 0.0.6 for example.
My content is version 0.0.2
Then it would apply migration 0.0.3 and 0.0.6
If content was 0.0.0
It would use all 3 migrations.
If I explained this properly, it should make sense that the "fromVersionRange" is not necessary. It's implicitly based on all migrations provided.

packages/plugins/content/slate/src/plugins/paragraph/node.js
PeterKottas: The way react decides if to update component is actually shallowEqual so it seemed redundant to do it twice. Unless there's something I am missing.

packages/plugins/content/slate/src/plugins/paragraph/node.js
PeterKottas: They removed the placeholder from slate. The placeholder is now provided as parameter to the editor.

arekkas:
> In the context of this library, I don't consider migrations to be "key" feature. Think of it this way, you have thousands of people using this lib, and so far it wasn't there at all. I am not saying this to force you to accept wrong solution to the problem, just pointing out it will probably not be used by too many people.

I disagree. Migrations of the document model are as important as migrations of a SQL Schema. If we significantly break the document model without a way for migrating to a newer version, it will be a terrible developer experience. The system was not in place because there was no need in the core library, but versioning existed from the beginning due to this.

Versioning for libraries is, unfortunately, more complex due to the way typical libraries are versioned (semver). This allows for high complexity when releases are not sequential (e.g. 0.9.0 released after 1.0.0 but 0.8.0 released before 1.0.0).

If you don't want to take time to think about this - which would be totally acceptable - I'll free up some time to do so.

PeterKottas:
(e.g. 0.9.0 released after 1.0.0 but 0.8.0 released before 1.0.0) is a very good point. Yeah I sort of agree with you. For now, I am tempted to leave it for another PR if that would work for you. The only thing I am afraid of right now is to wait a long time to merge this. If we can in fact merge it, I'd be happy to work with you on improving the migrations after this goes in.

arekkas:
> The only thing I am afraid of right now is to wait a long time to merge this. If we can in fact merge it, I'd be happy to work with you on improving the migrations after this goes in.

I created a 0.4.0 branch where we can merge this as work in progress. Once the changes are done we'll push it to master and tag it v0.4.0 . That way it's easier to work on this branch and also use it in npm. Deal?

PeterKottas:
Yeah sure, that works for me. So will you just merge this now, release it and create new PR against master? Did I get it right? The only problem then will be I won't be able to commit directly to it. But I can PR against it if that's preferable.

PeterKottas:
In the meantime, I think I might have cracked the migrations and do it properly this way.

Let's assume that all migrations necessary to migrate to current version are present in code. This is an important assumption and it should become guideline for developers. Let's call it assumption **A**

For the argument sake, let's create this example situation.

Existing plugin versions in order of releases

0.1.0 -> 0.2.0 -> 0.3.0 -> 1.0.0 -> 0.4.0 -> 1.1.0

We would then add the param you proposed so the migration object could look like:

```
export type Migration = {
    toVersion: string, // SEMVER, single version
    fromVersionRange: string // SEMVER range
    migrate: (state: Object) => Object
}
```
Sidenote, once/if we have a proper typesystem in place, we could also keep all state types and provide 2 generic arguments
```
export type Migration<FromState, ToState> = {
    toVersion: string, // SEMVER, single version
    fromVersionRange: string // SEMVER range
    migrate: (state: FromState) => ToState
}
``` 
Which would imho be pretty cool

Let's now create the migrations for our example

```
[
  {
    toVersion: '0.3.0',
    fromVersionRange: '0.1.0 – 0.2' // >=0.1.0 to < 0.3.0 according to https://devhints.io/semver
  },
  {
    toVersion: '1.0.0',
    fromVersionRange: '~0.3.0' // >=0.3.0 to < 0.4.0 according to https://devhints.io/semver
  },
  {
    toVersion: '0.4.0',
    fromVersionRange: '~1.0.0' // >=1.0.0 to < 1.1.0 according to https://devhints.io/semver
  },
  {
    toVersion: '1.1.0',
    fromVersionRange: '~0.4.0' // >=0.4.0 to < 0.5.0 according to https://devhints.io/semver
  }
]
```

Let's do this for content with version **0.1.0** and plugin version **1.1.0**

Pseudo code
```
**currentDataVersion** = '0.1.0'
**pluginVersion** = '1.1.0'
while **currentDataVersion** != **pluginVersion**
  **migration** = Find migration that upgrades from **currentDataVersion** based on **fromVersionRange**
  **state** = apply **migration**
  **currentDataVersion** = **migration**.**toVersion**
```

Based on **A** we can guarantee that after going through that while loop, even if we don't end up with exactly the same version as out **pluginVersion**, it should still mean we have correct data.

And that's pretty much it. Not complicated at all and if we agree on this, I can implement it later on today.

arekkas:
> Yeah sure, that works for me. So will you just merge this now, release it and create new PR against master? Did I get it right? The only problem then will be I won't be able to commit directly to it. But I can PR against it if that's preferable.

I will do another review pass, then merge

> And that's pretty much it. Not complicated at all and if we agree on this, I can implement it later on today.

Code example looks good. I think that's the way to do it!

PeterKottas:
I was trying to fix the lists, but weirdly, I noticed it doesn't work properly on the master either. Check out https://editor.ory.am/?utm_source=github&utm_medium=header&utm_campaign=editor, select one of the existing lists and notice that the button for list is not active. If you click it, it will wrap it once again.

The reason for this is it's wrapped in additional paragraph. Not sure how that happened. Do you ahve any insight?

arekkas:
I’ll check it tomorrow

PeterKottas:
I am not exactly sure how you suppress build errors like this one. What method you prefer/use?

arekkas:
We can ignore that for now!

PeterKottas:
Ok, fair enough ;)

PeterKottas:
Hey mate, is this not released yet? I am thinking maybe the flow errors broke the release pipeline?

aeneasr:
Yeah the lerna upgrade broke it, fixed now

PR #556, opened 2018-08-12
Fixed few issues

PeterKottas:
- Parallax bg input fields were not working
- If you didn't provide any content/layout plugins, instead of nothing, it showed "0" literally in the ui as that's the length of plugin array

### DIFF
--- a/packages/plugins/layout/parallax-background/src/index.js
+++ b/packages/plugins/layout/parallax-background/src/index.js
@@ -36,9 +36,9 @@ class PluginComponent extends Component {
   state = { hidden: false }
   props: LayoutPluginProps<{}> & { children: any }
 
-  handleChangeBackground = (e: Event) => this.props.onChange({ background: e.target.value })
+  handleChangeBackground = (e: React.ChangeEvent<HTMLInputElement>) => this.props.onChange({ background: e.target.value })
 
-  handleChangeDarken = (e: Event) => this.props.onChange({ darken: e.target.value })
+  handleChangeDarken = (e: React.ChangeEvent<HTMLInputElement>) => this.props.onChange({ darken: e.target.value })
 
   render() {
     const {
--- a/packages/plugins/layout/parallax-background/src/index.js
+++ b/packages/plugins/layout/parallax-background/src/index.js
@@ -32,11 +32,6 @@ import type {
 import { BottomToolbar } from 'ory-editor-ui'
 import ThemeProvider, { darkTheme } from 'ory-editor-ui/lib/ThemeProvider'
 
-const handleChange = (onChange: (state: any) => void, key: string) => (
-  e: Event,
-  value: string
-) => onChange({ [key]: value })
-
 class PluginComponent extends Component {
   state = { hidden: false }
   props: LayoutPluginProps<{}> & { children: any }
@@ -62,14 +57,14 @@ class PluginComponent extends Component {
               label="Image location (URL)"
               style={{ width: '256px' }}
               value={background}
-              onChange={handleChange(onChange, 'background')}
-            />
+              onChange={e => onChange({ background: e.target.value })}
+          />
             <TextField
               placeholder="0.3"
               label="Darken level"
               style={{ width: '256px' }}
               value={darken}
-              onChange={handleChange(onChange, 'darken')}
+              onChange={e => onChange({ darken: e.target.value })}
             />
           </BottomToolbar>
           {children}
--- a/packages/plugins/layout/parallax-background/src/index.js
+++ b/packages/plugins/layout/parallax-background/src/index.js
@@ -36,9 +36,9 @@ class PluginComponent extends Component {
   state = { hidden: false }
   props: LayoutPluginProps<{}> & { children: any }
 
-  handleChangeBackground = e => this.props.onChange({ background: e.target.value })
+  handleChangeBackground = (e: Event) => this.props.onChange({ background: e.target.value })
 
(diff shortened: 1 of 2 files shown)


PeterKottas:
Second issue can be easily seen if you navigate to https://editor.ory.am/?utm_source=github&utm_medium=header&utm_campaign=editor and search for "bla bla", you'll see "00", imho what you should rly see is a placeholder of some sort. "No plugins found" but for now this at least keeps it without bugs

PeterKottas:
Hell the no message on empty search for just stupid, I've added it right away.

arekkas:
Thanks! I am out of of office until Thursday and will review/merge then or earlier if I find some time. Cheers :)

PeterKottas:
Cheers, enjoy the time off, I have a feeling there will be a bunch of stuff waiting for you then :p

PeterKottas:
This bot is stupid!

arekkas:
DCO may fail if the sign-off name + email is different from the github one!

packages/plugins/layout/parallax-background/src/index.js
```diff
@@ -62,14 +57,14 @@ class PluginComponent extends Component {
               label="Image location (URL)"
               style={{ width: '256px' }}
               value={background}
-              onChange={handleChange(onChange, 'background')}
-            />
+              onChange={e => onChange({ background: e.target.value })}
```
arekkas: Defining functions in render methods is bad practice as it creates a new function on every render. Instead, please modify the `handleChange` function!

packages/ui/src/Toolbar/index.js
```diff
@@ -127,9 +133,9 @@ class Raw extends Component {
             )
           })}
         </List>}
-        {layout.length && <List subheader={
-            <ListSubheader>Layout plugins</ListSubheader>
-          }>
+        {layout.length > 0 && <List subheader={
```
arekkas: `content.length > 0` is the same as `boolean(content.length)`, no?

packages/plugins/layout/parallax-background/src/index.js
PeterKottas: Thanks that's actually right ... hmm I think I will have to look into this. Do you know how much this hinders performance? Cause I use this all the time

packages/ui/src/Toolbar/index.js
PeterKottas: yeah should be, I just prefer this syntax

packages/plugins/layout/parallax-background/src/index.js
arekkas: In case of this editor it's significant

packages/ui/src/Toolbar/index.js
arekkas: Let's keep it how it was :)

packages/plugins/layout/parallax-background/src/index.js
arekkas: The function should probably be:

```
const handleChange = (onChange: (state: any) => void, key: string) => (
   e: Event,
 ) => onChange({ [key]: e.target.Value })
```

Although `Event` should probably be SyntheticEvent or SyntheticInputEvent (not sure which one of the two without flow to check atm)

PeterKottas:
Sometimes the sign-off works, sometimes it doesn't. Driving me nuts. Always commiting the same way. But anyways, the anon functions are fixed, good catch

packages/ui/src/Toolbar/index.js
```diff
@@ -40,7 +40,8 @@ import Provider from '../Provider'
 
 type Props = {
   isInsertMode: boolean,
-  editor: Editor
+  editor: Editor,
+  pluginsNotFoundText: any
```
arekkas: This should be `string`, right?

PeterKottas: Not rly, normally i would do string | JSX.Element but I don't know how that works in flow. The thing is some folks might want to render some real jsx content. take "<h1>Test</h1>" as an example. Locking it down to string would restrict that

arekkas: Ah, I see, the right value would then be `Node`:

```
import type {Node} from 'react'

//...

  pluginsNotFoundText: Node

```

In that case I also want to propose to rename the property from `pluginsNotFoundText` to `noPluginFoundContent` or something similar - it should be clear that it can be text, an image, a react element, or whatever.

PeterKottas: Makes sense, fixed in latest commit

PeterKottas:
Apparently there's a problem in build with the Node type you suggest
```
import type { Node } from 'react'
                   ^^^^ Named import from module `react`. This module has no named export called `Node`.
```
Apparently that one doesn't exist. Is it a typo or can you suggest something else that would work pls?

arekkas:
Hm strange, might have to do with an outdated flow version. It's from the [docs](https://github.com/facebook/flow/blob/master/website/en/docs/react/types.md#toc-react-node):

> import type {Node} from 'react';

Let's have any then as a workaround for now

PeterKottas:
Ok fair enough. Yeah the version is truly archaic. Is this all for this PR then or should I fix something else?

arekkas:
No, that would be it. But `: any` is still missing :)

PeterKottas:
Haha, my bad. Too many changes flying around

arekkas:
Thank you!

PR #557, opened 2018-08-13
Implemented migrations

PeterKottas:
This PR implements migrations between versions. It's pretty straight forward. I though about having the migrations both ways, but tbh, I think it would never be used. How often would somebody decide to downgrade plugin version that was used for a while. 

Anyways, if at all possible, **could you treat this PR with higher priority** as it will likely be required for #554 especially if we don't want to introduce slate-next (which I think is something all of us want to avoid)


arekkas:
Thank you! I'm closing this PR because I reviewed the changes in #554 separately :)

Issue #555, opened 2018-08-12
New plugins ideas


PeterKottas:
I will be adding a few new plugins in the near future, I though this might be a good place to share ideas about it and possibly come up with some inspiration for features that would be appreciated by users. 
Let me first start with a question to @arekkas though.
Given the structure of the library (each plugin is a new package), is it reasonable to keep the growing number of plugins in this github repo? My fear is that build (and npm install) time will grow as we keep adding those. I know that you are open to 3rd parties but there are already a few plugins in this repo. Where do we draw the limit?

Anyways, here are the plugin ideas:

- colored-background (might also include gradients), this would obviously be layout plugin
- colored text - this is in fact a slate plugin that would be added so no big work necessary
- image background - again layout, simmilar to parallax background but without parallax effect (tbh, I would prefer to see a image-background plugin where you can turn on parallax rather than just parallax) also while improving, my colored-background might be stupid as well as this could be included here along with gradients
- image upload - unbelievable nobody shared this yet cause I am sure there's something out there (obviously there's some inherent complexity to this as it needs to be generic to allow for different providers/apis to work but it should be doable)
- divider with text - once again, either the old divider could be improved or new one can be implemented, expected end result to look something like this https://codepen.io/ericrasch/pen/Irlpm

PeterKottas:
Also one extra question, I would prefer writing new packages in typescript. The way flow is setup currently appears to have lot of problems and I am spending lot of time fighting those. Would this be a problem? I can obviously modify the build to work just fine.

arekkas:
I like all of those plugin ideas. They should definitely be in separate PRs. I think it's fine to have them in the main repository.

One issue with the repository set up is how watch mode works. It's painfully slow and resource intense as we're watching several directories independently. Adding another typesystem to the mix won't make it faster/easier to set up. I think we need to solve the flow issue first before we introduce typescript to this ecosystem.

PeterKottas:
Ok fair enough, in the meantime I am making all my efforts for new code in flow anyways. I guess it's just easier for the time being.

PR #560, opened 2018-08-16
Background plugin and image upload

PeterKottas:
Hi guys, this is all about the #555. Namely the background plugin. It's rather big so I figured a video would help you understand what it's all about, Enjoy! :) 🎥 https://screencast-o-matic.com/watch/cFjDlgq2HI 🎥


PeterKottas:
Cheers, just fixing some build issues (flow doesn't work very well for me so there are always problems) and it will be good to go.

arekkas:
> Unbelievably bad experience trying to fix these types problems :( I summed it up in #561 hope we can find a way to improve the situation cause right now it's quite hard to contribute big things (plugins and what have you)

Yes, the build system needs tons of love. I'll try to free up some time to do this.

PeterKottas:
I think this happens all the time. One year time in lifespan of build tools is like the whole world changed. I am just happy to have it done now and hopefully will find some time later to either go to typescript or at least fix this

arekkas:
I checked out the video, impressive stuff! I think we should address the other PRs first, then I'll give this tons of review love :)

PeterKottas:
I am not sure what's going on but these changes seems to have broken the example. Or I merged it wrong. I am getting 
```
Element ref was specified as a string (trigger) but no owner was set. This could happen for one of the following reasons:
1. You may be adding a ref to a functional component
2. You may be adding a ref to a component that was not created inside a component's render method
3. You have multiple copies of React loaded
```
Which I've never seen before

PeterKottas:
Would you please be able to pull down my background-plugin and try to run ```yarn start:watch``` on your end? There's something sinister going on.

arekkas:
Can you please run `yarn run clean && yarn run clean:lerna && yarn` and then `yarn start`?

PeterKottas:
Strange but that apparently helped. Also ```yarn start``` is actually watching for changes which I didn't think it did. Anyways all good now I think. Ready for review

packages/plugins/content/image/src/Component/Form/index.js
```diff
@@ -21,74 +21,106 @@
...
-        // noopener is safer but not supported in IE, so noreferrer adds some security
-        change.rel = 'noreferrer noopener'
+  handleChange = (onChange: Function) => (e: Event) => {
+    const target = e.target
+    if (target instanceof HTMLInputElement) {
+      const change = {}
+
+      if (target.name === 'target') {
+        if (target.checked) {
+          change.target = '_blank'
```
arekkas: `_blank` is considered [bad practice](https://css-tricks.com/use-target_blank/) and even invalid in some specs (strict XHTML). Can we remove this?

packages/ui/src/DisplayModeToggle/Button/index.css
```diff
@@ -28,7 +28,7 @@
 .ory-controls-mode-toggle-button-description {
   font-family: Roboto, sans-serif;
   font-size: 16px;
-  margin-top: 24px;
+  margin-top: 18px;
```
arekkas: Reason for this change?

PeterKottas:
These are really cool changes!

> Unfortunately, I disagree a bit with the direction of the parallax-background plugin. I think it should be split into two, a gradient-parallax-background plugin and the parallax-background plugin with image upload. I'm not sure if we should allow the parallax-background plugin to disable parallax? Maybe it would make more sense to have an image-background plugin which can do that? If we do move that into image-background I'd do it as a new plugin and deprecate parallax-background with a console.warn()

I agree with the splitting here, I actually though this might be a bit too much of change for just an upgrade. I would even be fine with leaving the paralax-bg as is and just adding this as a new alternative. Of course paralax-bg could be deprecated as this support it's features and more. But that's definitely optional. However I rly want to keep the ability to upload image and add gradient in the same plugin. The reason for that is you can easily make cool overlay effects which would be unachievable without it. 

> Regarding the padding, it looks really cool! The issue however is that users will just pick a value which looks good on their device but will not think about mobile, tablet, ... - IMO this feature should be removed because it will definitely be abused and will break content!

Here I agree 100%, I am actually a bit of a hypocrite, cause I rly didn't like this exact thing on the spacer plugin, that people can just choose arbitrary number for the space and go with it. It must just look weird most of the time. However what I would like to keep is the ability to disable the padding (just a simple switch). The main reason for this feature was to enable people to turn it off anyways. Because then, you can stack multiple plugins on top of each other once again creating advanced functionality and use cases. (well on top of image with color background for example)

packages/plugins/content/image/src/Component/Form/index.js
PeterKottas: This was actually there before. Not sure why it lights up as a change. If you prefer something else, we can change it, but I skimmed through that article and can't say I agree with it 100%. It all seems explained very well, but in some cases, I just want in to open in a new window. I must admit I didn't read the whole thing so maybe that's why.

packages/ui/src/DisplayModeToggle/Button/index.css
PeterKottas: I am pretty sure it was breaking something but couldn't remember what, give me a sec.

PeterKottas: Oh yes, I see it now, if you hover over the mode buttons. A little tool tip pops up. And it wasn't centered vertically, was driving me nuts :D So it's basically a fix. Unrelated to the plugin though, sorry about that. Unless it's intentional of course, then I can revert it.

PeterKottas:
Padding feature explanation: https://screencast-o-matic.com/watch/cFj3DUq3cn

packages/ui/src/DisplayModeToggle/Button/index.css
arekkas: Ok, let's keep it!

packages/plugins/content/image/src/Component/Form/index.js
arekkas: Oh sorry, totally my bad. Let's keep it then, and I'll punch myself in the face for adding it ;) The code shows up as a change because the formatting changed.

PeterKottas: Haha, no worries, yeah it does, I'll keep an eye on this for the future PRs. It's the vs-code autoformat feature that's doing this. I'll see if I can configure it further

arekkas:
Ah, that makes sense!

arekkas:
> The reason for that is you can easily make cool overlay effects which would be unachievable without it. Also agree with the fact that the naming would sort of break, if it's called parallax-bg it probably shouldn't be possible to disable parallax.

Ah, I thought you can't do both image and gradient at the same time, only one of the two? I haven't tested this branch yet locally (we really need a preview feature here...) but will do so tomorrow.

I like the idea of having this as a separate plugin, maybe just `background` or something?

packages/plugins/content/image/src/Component/Form/index.js
arekkas: Configure it to use `npm run prettier`!

PeterKottas:
There's more than meets the eye here my friend :p Feature branches deployments are rly cool! I configured it for one of my previous projects and it was rly a big success. Especially when you have to deal with all sorts of PM and other "important" people :) Yeah background sounds good to me. I'll reinstate the paralax-bg and add this as a new "background" plugin.

aeneasr:
I played around with the plugin a bit more. What I find confusing is that, when you set image, color, and gradient and switch back to image, you first have to disable all the other settings in order to see the image. It makes sense from a technical perspective (just how background css works) but it's really weird from the editing experience. Maybe exporting three plugins in the background package isn't such a bad idea after all? That would make it very obvious what you can use and how to remove it?

examples/src/index.js
```diff
@@ -40,7 +40,7 @@ import spacer from 'ory-editor-plugins-spacer'
 import 'ory-editor-plugins-spacer/lib/index.css'
 
 // The image plugin
-import image from 'ory-editor-plugins-image'
+import { imagePlugin } from 'ory-editor-plugins-image'
```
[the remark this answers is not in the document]
PeterKottas: ah cool, I was thinking you might be on holidays or something. afk now but I think this is not a breaking change. the only problem is the plugins now are exported as object instead of functions (the default export). that is generally not s good idea because it makes it impossible to pass in config. what I did here is i left the defaul export as an object (for backwards compatibility) and added the named export as a function so that you can configure it yourself. we could make a not of this in docs. but i think for the future, it would make sense to convert all plugins to be functions that return objects.

PeterKottas:
hmm, wouldn't want to export 3 plugins as there would be some code repetition. i actually find it more convininent w

PeterKottas:
(typing on new tablet and i somehow closed the pr by mistake. have to get used to this :) )

examples/src/index.js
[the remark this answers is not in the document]
aeneasr: Ahh, that makes sense - in that case no need for documenting this as a BC

aeneasr:
I don't think you need to repeat all three plugins. Just pass this as a property/config setting. E.g. (just some ideas):

```
import backgroundFactory from 'ory-editor-plugins-background'
backgroundFactory({ 
  defaultPlugin: slate(),
  type: 'image',
  imageUpload: fakeImageUploadService('/images/sea-bg.jpg')
})
```

You can also hide this config (do it in the exports of the plugin):


```
import {imageBackground, colorBackground} from 'ory-editor-plugins-background'
imageBackground({ 
  defaultPlugin: slate(),
})
colorBackground({})
// ...
```

This allows developers to really enable only those features they want. For myself, I really like the color and image plugin, but I definitely don't need the gradient one. So it would be great to be able to disable the last one.

aeneasr:
By disabling certain features / explicitly enabling it, we could also create three different plugins which each of the settings enabled stand-alone. So 1 would allow 2 :)

PeterKottas:
Yep that's right :) Ok, so 1 it is and it's already implemented. 
Summary of changes:
1. New settings on the plugin called *enabledModes* it accepts a number (flag). This flag sets up which modes are enabled. Example:
```
background({
  enabledModes: COLOR_MODE_FLAG | IMAGE_MODE_FLAG | GRADIENT_MODE_FLAG
}),
```
where the flags are exported as constants from the index of the plugin.
2. I've changed the UI from a switch to a Tab. (my wife is a lead designer and she told me very nicely if I am crazy when I am changing UI with a switch instead of tabs in material ui :D )

PeterKottas:
Btw, the pre-commit hooks seems to be causing some problems in vs-code sometimes. It caused some strange unexpected merge conflicts. Hard to explain what's going on, I'll keep an eye on it in case it repeats.

aeneasr:
This is awesome (especially the tabs!)!

PR #564, opened 2018-08-18
plugins/slate: Updates to latest slate and adds migrations (#554)

arekkas:


packages/core/src/service/plugin/index.js
```diff
@@ -162,6 +176,51 @@ export default class PluginService {
...
+        break
+      }
+      currentDataVersion = migration.toVersion
+      state = migration.migrate(state)
+    }
+    return state
+  }
+
+  getNewPluginState = (found: any, state: any, version: string) => {
+    if (!found.pluginWrongVersion || semver.lt(found.pluginWrongVersion.version, version)) {
```
arekkas: @PeterKottas if I read this correctly, the idea behind this is that, when calling `unserialize` with a specific plugin version, then `pluginWrongVersion` will be set if said plugin version does not exist in the `PluginService`, right? If so, a couple of questions:

1. Is the idea of `semver.lt(found.pluginWrongVersion.version, version))` here that, if we can not find right version, and the fallback plugin `*` is newer than the content version, we simply return the state with the `PluginMissing` plugin?
2. Does that imply that `pluginWrongVersion` is basically like an indicator that said plugin version was not found, but if a newer version was found we'll apply the migrations?
3. Assuming that we have two versions registered `0.0.2` and `0.0.3` and our content model is in `0.0.1`, would this work with this code? Seeing that you're returning the first best match to `*`?

arekkas:
@PeterKottas there is an issue remaining with lists which is that I am unable to nest lists using tab, or un-nest them with shift+tab. This works in the current demo (you must select the second list item as you can't nest the first list item). On this branch, it simply selects the toolbar at the bottom. I'm not sure which code change caused this but I think nesting lists is quite common editing pattern.

arekkas:
Apart from that and working a bit more on the PluginService I think this is pretty much good to go!

packages/core/src/service/plugin/index.js
```diff
@@ -162,6 +176,51 @@ export default class PluginService {
...
 
+  migratePluginState = (state: any, plugin: Plugin, dataVersion: string): ?Object => {
+    if (!plugin || !dataVersion || semver.valid(dataVersion) === null) {
+      return state
+    }
+    let currentDataVersion = dataVersion
+    let migrations = plugin.migrations ? plugin.migrations : []
+    while (true) {
+      const migration = migrations.find(m => semver.satisfies(currentDataVersion, m.fromVersionRange))
+      migrations = migrations.filter(m => !semver.satisfies(currentDataVersion, m.fromVersionRange))
```
arekkas: This does not seem to be used in the rest of the code @PeterKottas

PeterKottas: I think the code selection went bananas. Cause what I see mentioned is the migratePluginState function and that is definitely used. That's what migrate the state.

arekkas: My bad, I read the code wrong

PeterKottas:
Looking into the lists now, but frankly I am little bit lost. The only thing that comes to mind is maybe ... maybe ... material ui changed this somehow. Cause it's general practice to focus next clickable element when you click tab. And possibly they hijack the event somehow.

packages/core/src/service/plugin/index.js
PeterKottas: Ah cool, np

PeterKottas:
Ok I managed to fix the shortcuts. It was a mistake on my end. But there's still a problem with lists.
It's still missing the inner PARAGRAPH/PARAGRAPH that we spoke about earlier. As a result of that, the plugin is not working correctly. I will have to take another look

arekkas:
Thank you for the video! The `KEY_TAB` comes from `slate-edit-list`, right? It generally appears that there are issues with slate 0.37.7 and current `slate-edit-list` which appears to be on `^0.32.0`: https://github.com/GitbookIO/slate-edit-list/issues?q=is%3Aopen+is%3Aissue+label%3Aupgrade-slate

However, I pulled this plugin and updated to slate 0.37.7 and it seems to be setting `p` properly:

image

So it has to be an issue with the plugin. Not sure which one though.

arekkas:
This is definitely a state-issue. See:

image

The paragraph type is missing here.

While investigating I also noticed that `PluginService.schema` is no longer used (right?), so I think we can remove this field from the class and the plugins.

PeterKottas:
Yeah the KEY_TAB is defined here https://github.com/GitbookIO/slate-edit-list/blob/master/lib/index.js . I think the paragraphs might have disappeared because of the changes I made in the edit list plugin (ours not the lib for keyboard shortcuts). But frankly I am not sure what changes have caused it. The logic that implements it is here https://github.com/ory/editor/blob/0.4.0/packages/plugins/content/slate/src/plugins/lists.js#L60-L90 . I'll have another look but I can't say I understand the way changes work in slate 100%. Might have to head back to docs.

arekkas:
Thank you so much for your hard work. This looks all good to me! Merging.

PeterKottas:
Awesome!

PR #565, opened 2018-08-20
vendor: Move to npm and declare material-ui as peerDependency

arekkas:
This PR upgrades to latest material-ui and defines material-ui as a peerDependency of each module which removes significant vendoring bloat. This implies however that material-ui has to be a dependency in your project.

This PR should generally improve IDE performance and npm performance as significantly less files need to be scanned/downloaded.


PeterKottas:
I was about to report this as well! Nicely done.

arekkas:
Had to revert npm changes but was able to get rid of some other dependencies in the examples and use react-scripts only

nolandg:
I'm really confused. This commit seems to "update" material to 1.5 but latest version is 3.x. Is this a mistake or am I out to lunch? This editor seems awesome but will be very difficult to integrate if it's peer deps are 2 major versions behind.

aeneasr:
Relax ;) First of all, MUI maintainers moved from 1.5 to 3.0 directly (without 2.0 release), and the 3.x release isn't even 2 weeks old. I welcome you however to create a PR with the required changes to move to version 3.0

nolandg:
Ok, I worked on this and got all the errors and warnings to go away and the example working with latest MUI but I'm pulling out my hair trying to use the repo in local development to test it in my site.

I'm not familiar with lerna, mono repos, and local testing. I'm currently in a hell of recursive yarn links and unresolved modules. Eg: the example works great within the repo but when I try to use the same code in my app, it tries to pull in all sorts of wrong versions of peer deps from it's own node_modules.

I might eventually get the energy to properly learn the whole local test mono repo thing but if anyone wants to throw me a bone, I'll take it.

aeneasr:
Awesome to hear that you dedicated time for this! :)

I think the issue here is that we'll have to release it to npm first in order for you to consume it from the outside (or rather not from the internal example). I had some issues with npm as well, I think they are primarily caused because - maybe - npm does not resolve yarn.lock files from nested dependencies? This is also why we're currently still relying on yarn and not npm to properly fetch the right dependencies. I think the first step would be to upgrade to mui 3.x and then tackle npm as a separate thing.

nolandg:
As an example: https://github.com/ory/editor/blob/3ef3a273e3255e4a503695046a74ede39302d0eb/packages/plugins/content/image/src/Component/index.js#L28
Do you guys make a `yarn link` for every package that uses `<ThemeProvider>`? How else do you edit `<ThemeProvider>` code and see the effects on the packages that use it?

PeterKottas:
Dude you're reading way way too much into this. Just run npm run start. That's it. Lerna (monorepo manager) takes care of everything. You get the examples project up and running and have live reload and everything. Never would have though people would struggle with this but I guess we can put this in docs. Btw, I am afk so maybe you have to run start in example packages, not sure which one. It just pops up first in vscode :) And don't forget to install everything first.

nolandg:
That's what I started off doing but then Lerna installs all the devDeps from the root package.json and so my build includes multiple copies of MaterialUi etc which causes havoc--especially when they're different versions. Is there a Lerna command to start+watch but use the peer deps from the host application?  I've tried bootstrapping without hoisting and with nohoisting but it still includes its own copy of Material (and all other peer deps). Yes, I could publish the package and then it should use the host's peer deps but that's kinda of a long dev cycle.

How do you guys do it? Do you only work with the example package? or does your host app not use Material? Or am I missing the right lerna commands?

PR #567, opened 2018-08-20
Bump flow-bin and update typdefs 

arekkas:
@PeterKottas I updated flow-bin and improved on flow-types. Obviously there are errors now (~110) because it was a major bump and there are broken definitions in the code.

Unfortunately, there's also issues here. First is that despite my attempts to improve the flow check time it's still extremely slow. I ignored all node_modules (previously not the case), lib, and the examples. It's still very slow though.

Another issue is that flow isn't able to resolve the types from the packages that rely on one another. Instead we would have to write type definitions and commit them to flow-typed which is stupid and a common critique of flow. If I'm missing something here please let me know.

I think the time it would take to move to TypeScript would be the same as fixing everything in FlowType. Both will take significant amounts of time (~ 2-4 full days) unfortunately but I think TS has a better architecture, community, and ecosystem.

What do you say?


PeterKottas:
I very much agree with you here. My personal preference is typescript, I'll tell you that straight up. I have some experience with flow but very often encountered similar problems to the ones you are describing. With TS, it just works. The speed is not an issue there as well. I recommend using https://github.com/s-panferov/awesome-typescript-loader to compile to JS which has superior performance to combination of typescript and babel. (it's doing some caching behind the curtain) You can also check some of my libs, for example https://github.com/PeterKottas/guestbell-forms which uses typescript and builds to bundled minified js quite successfully. 
Overall Typescript imho is more supported, faster, has better support in IDE (vscode at least MS love MS I guess) Support for libs is also quite good. Plus if we write this in typescript, there's an easy way to generate d.ts and include it in the bundle so people can have typesystem in the minified lib for free. I use this lib to do that. this is the one https://github.com/TypeStrong/dts-bundle
Hard to predict how long will the "rewrite" take though. The code is not bad, there just might be some unexpected problems. I would say 1-6 days depending on the amount of unexpected things to fix.

arekkas:
Cool, I agree with this and the advantages seem to be plentiful. I've never used TS in a project, I would like to keep it as standard or "zero-config" as possible, just like react-scripts I guess. Reason being that upgrading is just easier.

dts-bundle looks like a perfect choice for the issue we're facing (consuming packages from other packages within this repo). If we run into performance issues I'd be open to check out awesome-typescript-loader too!

I'm not sure how to work on this. I think the biggest packages are core and slate, everything else is peanuts. I'd suggest to start with core as everything else depends on it. I'm not sure if I can dedicate another day this week to that but if I find the time I'll see how far I can get.

arekkas:
I'm merging this to get the changes into 0.4.0, feel free to further discuss here.

PeterKottas:
Ok sounds good. For me, right now it makes most sense to merge in everything that I am up to. When it stabilizes, we can start talking about the rewrite. Moving to typescript would benefit us as well as we could consume this library along with type system which would be great. (trying to have that luxury with every lib we use)

arekkas:
I'll try to introduce TS to the toolchain today for a small plugin (spacer) and set up a PR

arekkas:
Is it possible that babel-preset-typescript is not compatible with babel v6.0.0 and thus react-scripts? If so, did you find a workaround?

PeterKottas:
Never heard of this. I always just used awesome-typescript-loader

arekkas:
awesome-typescript-loader is a loader for webpack, not a babel preset/plugin :) We're not using webpack in the libs

PeterKottas:
Ah didn't know that. Never looked into the build pipeline but now it makes sense. The only major annoying thing about typescript is sometimes (if you are not using webpack) the tools tend to build ts=>js in place. So you end up with bunch of generated files in the workplace. It's always avoidable via settings but it cough me a few times depending on what tools you use.

arekkas:
#568

arekkas:
basically not possible right now, maybe in a month or two

arekkas:
or it's possible but it's a ton of work just getting the build chain to work, and it will probably break with the next beta/rc so not worth the effort atm

PeterKottas:
I don't rly mind the wait. But a question pops in right away, why don't you just go with webpack? I know you don't use it right now but would it be the end of the world to start using it? The only possible drawback is I have no idea how exactly would it work with lerna. But other than that, I build my frontend libs with webpack and I was never unhappy about it.

PeterKottas:
Imho, it's not so uncommon to use bundler for libs. At the end of the day, it's just about creating consumable output out of the code. Transpiler and bundler both achieve something along those lines. But where I agree with you is lerna would probably have to spin up NUM_OF_PACKAGES webpack instances which sounds ridiculous. Another option would be to run tsc first. Get ts -> js and then run babel on the js just to transpile es features. But again, this might hinder performance. If what you're sayin is that babel 7 will support typescript and achieve what I previously described in one step. Then it totally makes sense to wait.

arekkas:
What I tried to say is that with webpack, we'd still use babel. We could use the suggested webpack plugin instead (I think they use tsc to pre-compile, then babel?) but I'd prefer to have these things all in babel.

Webpack performance is quite terrible in a monorepo and I remember tons of issues from rebuilding to bundling dependencies to how to deal with css and which module output to use and also huge issues with react's singleton...it was a huge mess

PeterKottas:
Ok I think that sums it up perfectly! Hope the tools will get mature enough to be used in a stable within the next few months.

PR #570, opened 2018-08-21
Fixed lists

PeterKottas:


PeterKottas:
Getting tons of merge conflicts here :( Apparently you are using different formatting rules because the whole thing lights up as conflict. I managed to fix the lists to a working state. Also added buttons to increase and decrease indent. I know this is somewhat opinionated but I myself had no idea how to do it with shortcut. So this should hopefully make life easier for people. The relevant change is in https://github.com/ory/editor/pull/570/files#diff-526c2546141374aa41580dbbeac4f826
Notice how we were reinventing the wheel when the edit-list-plugin actually does all the heavy work for us. Now it works like a charm.

arekkas:
Formatting is standardized with `npm run prettier`.

PeterKottas:
Ok I'll give that a go. Tbh, I probably should have cloned 0.4.0 and worked on that. But I just continued on my master and it backfired. Anyways, the code now looks good.

arekkas:
Never work on master directly, especially with forks. Always create feature branches. I use https://github.com/Originate/git-town to make this easy!

PeterKottas:
Sure, generally it's never the case. Just sometimes, when contributing via fork, it's tempting to just fork, change stuff and PR back in. Thanks for the tip man! I'll definitely check it out. I am using git kraken lately which is awesome imho. (yeah I am a UI guy :D , if it's done well, I prefer it to console )

Issue #580, opened 2018-08-30
Make plugin controls and render components replaceable 


PeterKottas:
Let's first talk about this. 
Right now I am working on the client app, place where the content is rendered but not modified. Problem is right now I am forced to install material-ui which I would like to avoid. It seems that in order for the HTML (and later react-native) renderer, this lib is not required but as both serialize/deserialize and the rest of UI is in one file, you need it even when it's not used. My thinking is we could separate it away into another file (renderer.js for example) then the plugin index.js could load and include this but you would also have an option to just load the bare-bones necessary for the rendering. It sounds simple but it seems like I might be forgetting about something, plus I am doing this AFK so I am almost sure I definitely am. Let me know your thoughts. (also this would make implementing RN much simpler)

aeneasr:
Why is it a problem to have it in the dependencies? Maybe some plugin actually needs MUI for rendering! I'm not sure how this would work as a separate file, depending on how the plugin is packaged you might just get a single file (bundle).

aeneasr:
Your plugin can export `StaticComponent` alongside `Component` which is the component to be used for (HTML) rendering. I'm not sure if we can avoid `Component` from being loaded completely, but shouldn't something like tree-shaking do that for us?

PeterKottas:
Yeah that's true but like you've said, it would still include that in the bundle. We would have to load it dynamically in order for the tree shaking to work. I have some other solution in mind. Will get my hands dirty and put together a POC. I think that's the best way to explain what I am after. Also the way I want to do it will allow for RN implementation later.

aeneasr:
Hey the video was great for understanding what you're up to. This concept is really amazing, I love it! This will greatly improve the API of the plugins and again we see how react shines here!

I think it would even be an option to have the editor-ui renamed to `ory-editor-mui-ui` which ships UI components for the default plugins. This will make it a breaking change but it will also significantly improve the packaging. This is awesome!

aeneasr:
Once we have typescript, it will be a no-brainer to couple mui-ui with the typedefs of the plugins and maybe even add some tests or whatever. This is really great, excited to see this!

PeterKottas:
[...]

PR #587, opened 2018-09-03
Separate renderer and controls for image plugin

PeterKottas:
New PR that adds separation of UI and Rendering for the plugin. This is functionality further discussed in #580 . Please note this is a dummy PR that mostly serve as POC and will probably either be modified or abandoned in favor of more complete solution in the future.


aeneasr:
This looks good to me! There's still some issues I see. First is that HTMLRenderer still relies on mui. Second is how we can initialize the plugins properly. I was thinking that we could do something along the lines of:

```
import { Icon, Controls } from 'ory-editor/image/mui'
import image from 'ory-editor/image'

const e = new Editor({
  plugins: {
    content: [image({ Controls, Icon })]
  }
})
```

To make this easier, we could probably also do something like this

```
import { plugin as image } from 'ory-editor/image/mui'

const e = new Editor({
  plugins: {
    content: [image()]
  }
})
```

which would basically set everything up with the mui defaults.

Another idea I had is to first introduce `@ory-editor/*` or maybe `@ory/editor-*` and have two plugins, like `@ory-editor/image` and `@ory-editor/image-mui` but I'm not sure if that really improves things.

What do you think?

PeterKottas:
Yeah it still needs some polishing. I had some ideas as well but first need to finish some other stuff. The renderer definitely shouldn't rely on any UI packages. Unless of course intended by the plugin author.

Right now I am thinking that the plugins should:
1. Only keep core logic.
2. Controls and rendering should be different packages (to keep things lightweight), however I wouldn't be totally opposed to keeping some defaults in the plugin package. Main reason is contributors would probably prefer maintaining just one package instead of minimum 3 per plugin.
3. Plugin should always return class that extends PluginBase (I was first thinking it should return function, that would still work, but read bellow)
4. Plugin base should
a) Accept plugin config (generic type that extends defaultPluginConfig, that holds stuff like plugin name, description)
b) Implement the basic settings overides (like discussed in previous point)
c) hold previewState and expose onChange, and onPreviewChange (previewState is important because sometimes you need to change plugin state, but not update editor state (because it would fuck up undo/redo), think slider that would flush 10s of useless undos into the editor). I actually implemented something similar in background-plugin, but I think this should come out of the box. 
d) Handle switching between rendering and controls (basically a base render function), it would also inject the previewState and state into the renderer and controls. Basically just some out of the box functionality to make things easier. This would also enforce people not to render any html specific tags because that would destroy react-native capabilities.
e) the base render function should accept one parameter, controlsApi. The plugin implementation would provide this and the base render would inject it into controls. 

I don't want to go more in-depth, some of these will require POC, and I see some problems but overall that seems like a sensible approach to me. This should also be very well documented to make it easy for people to contribute.

Things I like about this:
1. Just one class and some functions would implement all core logic.
2. 2 sfc (because state is kept in the base plugin) required to implement renderer and controls
3. People could easily provide different controls and or renderers for existing plugins

> Another idea I had is to first introduce @ory-editor/* or maybe @ory/editor-* and have two plugins, like @ory-editor/image and @ory-editor/image-mui but I'm not sure if that really improves things.

I like the idea, I would maybe do @ory-editor/image, @ory-editor/image-controls-mui, @ory-editor/image-renderer-html

aeneasr:
The idea of separating the renderer from the image is that you might add other renderers later on?

PeterKottas:
Yep exactly. I thing supporting react-native out-of-the-box(ish) can rly make ory editor a superstar on the field of content editors. People (imho) when using react-native will only need to render data, not necessarily change it (controls) and if we allow them to do it, that's a big game changer. 

Imagine, if you wanted to support image plugin on RN now, you would basically have to recreate the whole plugin functionality. With separate renderer, that's literally <10 lines of code.  Which can even be released as a package.

And as we will eventually want this for GuestBell (my company), I will probably provide these renderers for most stuff. And then we can just put the magical words "Supports react-native" in big letters in editor readme ;)

PeterKottas:
Plus on top of this, you can even provide different renderers for html. Maybe image can have number of different renderers somebody provides where one adds some nice frame around it ... or something like that. Then we can change this dynamically on the example page, and people can check and download the one they like the most. (all this is theory but you see my point)

aeneasr:
Yeah that makes sense!

PR #601, opened 2018-09-28
layout/background: Only render BottomToolbar when we are in edit mode

caresx:
Let me know if I forgot to run any build scripts.

I'm also interested in adding i18n to the project, would that be accepted?

### DIFF
--- a/packages/plugins/layout/background/src/index.js
+++ b/packages/plugins/layout/background/src/index.js
@@ -375,6 +375,7 @@ class PluginComponent extends Component {
     const {
       children,
       focused,
+      readOnly,
       state: {
         darken = this.props.defaultDarken,
         lighten = this.props.defaultLighten,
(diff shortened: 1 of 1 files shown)


aeneasr:
Could you give a short description of what this solves? And please sign your commit with `git commit --amend -s` so the DCO is signed.

i18n is interesting but it would require some good amount of planning before being approached!

packages/plugins/layout/background/src/index.js
```diff
@@ -401,104 +402,105 @@ class PluginComponent extends Component {
               backgroundImage: `linear-gradient(rgba(0, 0, 0, ${darkenFinal}), rgba(0, 0, 0, ${darkenFinal})),linear-gradient(rgba(255, 255, 255, ${lightenFinal}), rgba(255, 255, 255, ${lightenFinal}))`
             }}
           />
-          <BottomToolbar open={focused} theme={darkTheme}>
```
aeneasr: I think it would be enough to have `open={focused && !readOnly}` here

aeneasr:
> The commit should be signed, I'm not sure why DCO says it's not. If you go to commits it says verified with my gpg key.

Right, it's usually when nickname or email don't match from the git commit with your github identity. No biggie :)

Thank you for the explanation, I added a comment.

packages/plugins/layout/background/src/index.js
caresx: I'm using the style used elsewhere throughout the codebase (e.g. https://github.com/ory/editor/blob/master/packages/plugins/content/html5-video/src/Component/index.js#L47), which I think makes more sense too. I could wrap it in parenthesis though.

aeneasr: Ok, makes sense :)

aeneasr:
Thank you!

PR #604, opened 2018-10-03
npm: explicitly set packages as public for public npm organization pu…

colepacak:
…blishing

Signed-off-by: colepacak <cole@favish.com>

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,5 +40,8 @@
     "reselect": "^3.0.0",
     "semver": "^5.3.0",
     "uuid": "^3.0.1"
+  },
+  "publishConfig": {
+    "access": "public"
   }
 }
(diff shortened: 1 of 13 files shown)


colepacak:
We're embracing this project as an inclusion into a Drupal module. For the short term, we've forked the repo, but we'd like to upstream any improvements that would be helpful in a general way for you guys. These changes enable us to group packages publicly under an NPM organization, which is the default for regular NPM accounts.

aeneasr:
Can you please point to the documentation where this setting is specified? I can not find it in the official npm docs.

colepacak:
We had to piece this one together, so bear with me.

Packages can be scoped to an organization (e.g. material-ui).
_https://www.npmjs.com/docs/orgs/publishing-an-org-scoped-package.html_

"When publishing scoped packages, the access level defaults to restricted. If you want your scoped package to be publicly viewable (and installable) set --access=public"
_https://docs.npmjs.com/misc/config#access_
_https://www.npmjs.com/docs/orgs/publishing-an-org-scoped-package.html_

publishConfig sets "config values that will be used at publish-time. It's especially handy if you want to set ... access".
_https://docs.npmjs.com/files/package.json#publishconfig_

aeneasr:
Thank you for the clarification!

PR #616, opened 2018-10-29
Adds ability to initialize Editor with your own redux store

### LINKED ISSUE #611: [Feat request] Ability to use existing redux store
We use redux in our project too, and unfortunately this means that both redux stores have the same `storeKey` (see https://github.com/reduxjs/react-redux/blob/master/docs/api.md#arguments). This means that our widgets which try to `connect` to our store from inside `ory-editor` can't connect, as the `storeKey` is the same. We have renamed our `storeKey`, but unfortunately many popular `redux` libraries such as `redux-form` don't currently support passing a custom `storeKey` (other than 'store'). 

I understand that `redux` is used internally by `ory-editor` and that you probably don't want it to be a `peerDependency`, but ideally 

a) you use a different `storeKey` internally in this project, to avoid namespace collisions with implementers
b) its possible to use this library with an existing redux `store` under a specific key (the way many redux-based libraries handle this case)

Thanks for reading!

davidfurlong:
Before I complete the PR with the stated TODOs below, are you happy with the decisions I made to implement this feature?

Now you can do 
```js
import { oryReducer } from 'ory-editor/core'

combineReducers({ 
  ory: oryReducer, // must be 'ory' key
  // rest of your reducers
})

// create the store using your reducer

new Editor({ store: store })
```
Fixes #611 

Todos
- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -30,10 +30,11 @@ import { selectors } from './selector'
 import PluginService from './service/plugin'
 import pluginDefault from './service/plugin/default'
 import type { Editable as EditableType } from './types/editable'
-import type Store from './types/redux'
+import type { Store } from './types/redux'
 import forEach from 'ramda/src/forEach'
 import HTML5Backend, { NativeTypes } from 'react-dnd-html5-backend'
 import { DragDropContext as dragDropContext } from 'react-dnd'
+import reducer from './reducer';
 
 let instance: Editor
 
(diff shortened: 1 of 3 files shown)


packages/core/src/index.js
```diff
@@ -135,7 +140,7 @@ class Editor {
 
   setContentPlugins = (plugins: Array<any> = []) => {
     this.plugins.setContentPlugins(plugins)
-    console.log(this.store.getState())
+    console.log(this.getState())
```
aeneasr: This can be removed

packages/core/src/index.js
```diff
@@ -153,7 +158,7 @@ class Editor {
   query = {}
 }
 
-export { PluginService, Editable, Editor }
+export { PluginService, Editable, Editor, reducer as oryReducer }
```
aeneasr: Just export it as `reducer`

packages/core/src/index.js
```diff
@@ -82,26 +83,30 @@ class Editor {
...
         'You defined multiple instances of the Editor class, this can cause problems.'
       )
     }
-
     instance = this
-    this.store = createStore(initialState(), middleware)
+
+    this.store = store || createStore(initialState(), middleware)
+    // If you use your own store, the store will be combined under the key 'ory'
+    this.getState = store ? () => this.store.getState().ory : this.store.getState
```
aeneasr: I'm not sure if this is a good idea, it could mean that many selectors down the road (e.g. for cells) could break. I'm not sure if it does, but it might very well be the case. What's the downside of just merging the two states?

packages/core/src/index.js
davidfurlong: Sure, theres other console.logs that probs shouldn't be there either.

packages/core/src/index.js
davidfurlong: Yeah this is main thing I wanted feedback on. Without this the end user would have to have 
```
  editables,
  display,
  focus,
  settings
```
as top level redux-store keys in their store. If there was only one such key and it had a sensible name this would be alright. Alternatively a compromise could be adding an `ory_` prefix to each of these keys. (Or nesting all these keys within an `ory` key - potentially a better solution). Having those keys as top level `store` `keys` is not a particularly attractive option.

However I may be missing a better way to compose reducers. I agree that my PR solution is not ideal in terms of future selectors.

aeneasr: Yeah I see how that makes sense with an existing store. I guess encapsulating this in a subkey is the way to go then, but it *may* won't work very well with existing selectors. Have you tried if everything still works as it should? So moving cells, editing content, .... when a store is set?

davidfurlong:
I will put it in a subkey in this PR then and test to make sure it works
with existing selectors

aeneasr:
Nice!

PR #624, opened 2018-11-20
Image bg css fix

PeterKottas:
This was causing the backstretch to span whole container. 

Signed-off-by: Peter Kottas <petokottas@gmail.com>


PeterKottas:
Thanks mate. Yeah it's been a while :) I might add some more stuff in December. But been quite busy lately. All of the things we discussed are still in the pipeline though.

PR #626, opened 2018-12-08
Bugfixes in html5video

dimofte:
A few things were broken:
1. the `<TextField>` wasn't changing its value
2. the width was very small (I changed to be the same as in the other `video` plugin)
3. when the value was changed, the html element wasn't updated accordingly (needed a page refresh to show the new content)

### DIFF
--- a/packages/plugins/content/html5-video/src/Component/index.js
+++ b/packages/plugins/content/html5-video/src/Component/index.js
@@ -29,8 +29,8 @@ import { darkTheme } from 'ory-editor-ui/lib/ThemeProvider'
 
 export type Props = ContentPluginProps<*>
 
-const changeUrl = (onChange: any) => (event: Event, newValue: string) =>
-  onChange({ url: newValue })
+const changeUrl = (onChange: any) => (event: Event) =>
+  event.target && onChange({ url: event.target.value })
 
 const HTML5Video = ({
   readOnly,
--- a/packages/plugins/content/html5-video/src/Component/index.js
+++ b/packages/plugins/content/html5-video/src/Component/index.js
@@ -51,6 +51,7 @@ const HTML5Video = ({
           label="Video url"
           onChange={changeUrl(onChange)}
           value={url}
+          style={{ width: '512px' }}
         />
       </BottomToolbar>
     ) : null}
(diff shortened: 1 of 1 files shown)


packages/plugins/content/html5-video/src/Component/index.js
```diff
@@ -51,10 +51,11 @@ const HTML5Video = ({
           label="Video url"
           onChange={changeUrl(onChange)}
           value={url}
+          style={{ width: '512px' }}
```
aeneasr: Maybe have that as minWidth? It would be even better if that would be a responsive size as we could view the video on even smaller screens

aeneasr:
Thank you! I've added a comment on the width :)

packages/plugins/content/html5-video/src/Component/index.js
dimofte: Thanks for the feedback!
I did think about using relative widths (e.g. 80%), somewhat consistent with ory's cells. But in the end, the scope of this PR was just fixing urgent matters, so I reused the exact style of the other plugins' form (`video` and `image`).
Imo, decisions to change the forms' behaviour should be taken to a different PR which would alter all plugins consistently

aeneasr: Hm, but the video image uses relative width: https://github.com/aeneasr/ory-editor/blob/master/packages/plugins/content/video/src/Component/Display/index.js#L35-L52

aeneasr: Same for the [image plugin](https://github.com/aeneasr/ory-editor/blob/master/packages/plugins/content/image/src/Component/Display/index.js#L33-L49) as well as [here](https://github.com/aeneasr/ory-editor/blob/master/packages/plugins/content/image/src/Component/index.css#L23-L31)

dimofte: But that's a different thing: this width is the form's, not the player's.
Here's the relevant code for the [image plugin](https://github.com/aeneasr/ory-editor/blob/master/packages/plugins/content/image/src/Component/Form/index.js#L114) and the [video plugin](https://github.com/aeneasr/ory-editor/blob/master/packages/plugins/content/video/src/Component/Form/index.js#L47)

aeneasr: omg yeah, I completely overlooked that. Sorry! :)

aeneasr:
Thank you!

packages/plugins/content/html5-video/src/Component/index.js
dimofte: yeah, I got that, no worries :)

PR #628, opened 2018-12-09
Updated babel, react-scripts and material-ui + deprecation warnings

PeterKottas:
Just like the summary says,
updated to babel7, latest react-scripts, @material-ui/core and @material-ui/icons (this surfaced a problem with react-jss which I fixed). Also cleared som deprecation warnings.


PeterKottas:
Weird error, any ideas what this is about? Maybe ci config needs update for node version?

aeneasr:
Yeah, the CI is using node version 6 while we require >= 8. Probably just bump it to 11? https://github.com/aeneasr/ory-editor/blob/master/.travis.yml#L4

PeterKottas:
🎉

aeneasr:
Awesome, I'll pull that and check it out locally but already looks pretty awesome

PeterKottas:
Cheers, I'll be updating react-dnd as well. But seems to be a little bit more work (missed a few major versions there :) ) so better put this forward first and I'll branch off it later.

aeneasr:
Seems like babel-core 6 is still referenced somewhere:

```
[1/5] Resolving packages...
warning Resolution field "babel-core@7.0.0-bridge.0" is incompatible with requested version "babel-core@^6.0.0"
warning Resolution field "babel-core@7.0.0-bridge.0" is incompatible with requested version "babel-core@^6.0.0"
[2/5] Fetching packages...
```

Is that a problem?

PeterKottas:
Good catch, I forgot to remove the "resolutions" in main package.json. It's not required anymore.

PeterKottas:
Just as a preemptive strike to questions you might possibly have. The reason why I ditched .babelrc nad went with js config is simply there are now some known bugs in jest that forces you to do so. Reason why there's babel.config.js in every package now is babel changed resolution and not it doesn't go up folders until it finds one, it simply stops at the first folder that contains package.json. So that was a necessary change. I was forced to ignore hover tests, because the way they are written simply doesn't work anymore with jest. These tests will have to be rewritten and it makes sense to do that once we move to typescript. (this PR enables us to finally do that)

PeterKottas:
Update: I was inpatient, branched of this and started working on typescript. And ... it works :) And rocks 🎉 Also it seems like it will be less work than I initially though. Starting now. We'll see how it goes.

aeneasr:
Shieeeeet!

PeterKottas:
:) this is the commit for divider. https://github.com/PeterKottas/editor/commit/d1267e160cbb73f84649f42c4849c5aa27277680 as you can see, it's pretty straight forward. That being said, core will be fun.

aeneasr:
LGTM, thank you!

Issue #598, opened 2018-09-19
Slate deprecation warning


davidfurlong:
Since the newest 0.4.4 versions of the ory slate plugin, my console is showing ```Deprecation (0.37.0): The `range.startKey` property has been deprecated, use `range.start.key` instead.``` in  slate-dev-logger.es.js

PeterKottas:
This is tracked here #571 . Unfortunately that warning doesn't come from our code but from the list plugin. They have a PR going in to fix that. Haven't checked for a while though.

aeneasr:
Right, thanks!

PR #629, opened 2018-12-14
Typescript rewrite

PeterKottas:
🎉 It's alive 🎉 

Highlights of this PR:

- Typescript rewrite done
- All tests working
- Prettier runs by reading tslint config.
- f12 navigates between projects properly
- Fixed a gazilion weird little nuances that had to do with types, typos or problems in code.
- This also fixes a lot of existing comments that were pointing to various issues. Most of these had something to do with es-link or flow types in general
- Changed tooling section in readme
- Added a Known issue in readme (sometimes vscode locks declaration files so you have to run build again)
- Removed all traces of flow and generally cleaned up the code
- Declarations are exported along with js files and js.maps. That means typescript people will love us.

Admittedly, this is probably biggest PR in the history of universe :D I rly tried to check everything to make your life simple. Hoping it will go as smooth as possible. Also I am here for any questions you might have.

PR #630, opened 2018-12-17
Plugin controls renderers

PeterKottas:
Also only pay attention to this pr after #629 lands as it's branched of that.

Plugins reworked. As described in #580.

Highlights of this PR
Ability to replace renderer and/or controls for plugins.
No changes to core.
Backwards compatible.
Fully supports typescript which makes it easy to work with.
It's an implementation of more powerful guideline instead of reworking how plugins works.


PeterKottas:
Change in plugin philosophy. Right now, most plugins commit the state to ory state immediately as you update their control values. Sometimes it makes sense, but in some cases this causes problem. For example spacer, if you scroll real fast, not only does the movement get jerky cause it's drowning in events, it also spawns undo/redo with hundreds state making it hard to work with. I've changed this by implementing something I call preview controls. (once again, not a change in core, just philosophy). Basically, plugin keeps a copy of state (or it's part in it) and only commits the state change when certain criteria is met. (slider is released, certain amount of time has passed, and so on) Just though you should know ;) Should make it easier to review.

Also note that this release updates slate to latest version and therefore fixes #598 and probably some other issues. As a result I had to create https://github.com/guestbell/slate-plugins which holds implementation of advanced slate plugins. (Gitbook abandoned original slate version so this was the only way to move forward)

aeneasr:
Since the other PR was squash-merged we have some conflicts here now. One possibility is to go the `git diff` route and create a new PR based off master, which shouldn't take more than ~30 secs. Let me know if you need help with that.

PeterKottas:
I tried merging master into my, but there's some problem with it. Not sure what this squash means but apparently it forgets all commits and pretends it's just latest. How should I do this now?

PR #634, opened 2018-12-23
Plugisnss rewrite

PeterKottas:
Hopefully finally closing #633 (and making me hate git more every day)


PeterKottas:
Cheers man. Yeah this got a big crazy :D It's not unreasonable what you're asking though. Just git doing unexpected stuff. I don't remember the exact commands but it was copy paste from what the bot told me to do. Unfortunately, I think the force push forced it to think it's totally different from master.

PR #638, opened 2019-01-17
Bump video package version to 1.8.0 to fix SSR

### LINKED ISSUE #637: Video plugin has a SSR issue
We had a really strange issue with the server side render of Ory renders related to having the video plugin included as part of the Editor plugins, but it wasn't being used in the rendered page. It was malforming and jumbling `div`s outside of the rendered React Component, applying styles to them and generally seemingly mutating the DOM outside of `<HTMLRenderer/>`

I can take a look at what the issue might be, but for the moment I wanted to document the issue here.

davidfurlong:
version 1.7.1 of react-player introduced a SSR bug by referencing navigator in global scope without checking whether it was undefined first. This was fixed by 1.8.0. Fixes #637

### DIFF
--- a/packages/plugins/content/video/package-lock.json
+++ b/packages/plugins/content/video/package-lock.json
@@ -42,9 +42,9 @@
 			}
 		},
 		"react-player": {
-			"version": "1.7.1",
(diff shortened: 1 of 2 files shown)


davidfurlong:
I don't think this was actually the issue breaking SSR.

PR #639, opened 2019-01-19
updated code to catch potentially inconsistent state updates

imsolost:
Hi! I found this [alert on LGTM.com](https://lgtm.com/projects/g/aeneasr/ory-editor/alerts/?mode=tree) for your project. This may not be the type of contribution you normally receive, but I figured it help clean up the codebase.

(Full disclosure: I'm part of the team behind LGTM.com)

Let me know if you have any questions.

### DIFF
--- a/packages/core/src/components/Cell/Resizable/index.tsx
+++ b/packages/core/src/components/Cell/Resizable/index.tsx
@@ -63,7 +63,9 @@ class Resizable extends React.PureComponent<ResizableProps, ResizableState> {
     }
 
     this.props.onChange(newSize);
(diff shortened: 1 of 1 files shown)


aeneasr:
Sorry, unless there is an issue with this (don't think so) I don't think it makes sense to merge.

imsolost:
No worries. It is a minor issue for sure, but a React component state updates using setState may asynchronously update this.props and this.state, thus it is not safe to use either of the two when calculating the new state passed to setState. The code will usually work fine without this change, but it can sometimes lead to unstable builds.

PR #583, opened 2018-08-31
Update Decorator.js to handle searchNodeEverywhere returning null

davidfurlong:

PR #643, opened 2019-02-02
Check if node exists on focusNext and focusPrev

### LINKED ISSUE #579: searchNodeEverywhere() is null
Sentry is showing two of errors which frequently occur from our implementation of ory-editor:
`Cannot read property 'node' of null` and
`r.props.searchNodeEverywhere(...) is null`

They both reference https://github.com/ory/editor/blob/cc6ba2bb1e4b4f65649fd1928cdc8ce756732b66/packages/core/src/components/HotKey/Decorator.js#L135 when `this.props.searchNodeEverywhere(focus)` don't find anything and this expression is evaluated as `null`. The function definition of searchNodeE verywhere has `null` as part of its valid return types, https://github.com/ory/editor/blob/cc6ba2bb1e4b4f65649fd1928cdc8ce756732b66/packages/core/src/selector/editable/index.js#L91 so I'm assuming that the first place (the only place in the codebase where we call `searchNodeEverywhere`) we need to check for a null value.

fernandobandeira:
#583 fixed #579 but I'm still receiving those errors on sentry sometimes:
`null is not an object (evaluating 't.props.searchNodeEverywhere(t.props.focus)`
`Cannot read property 'node' of null`

I've seen that they are coming from the focusNext function, I'm simply applying the same fix that was applied to the remove function to the focusNext and focusPrev function.

### DIFF
--- a/packages/core/src/components/HotKey/Decorator.tsx
+++ b/packages/core/src/components/HotKey/Decorator.tsx
@@ -129,7 +129,11 @@ class Decorator extends Component {
         return;
       }
 
-      const { node: n } = this.props.searchNodeEverywhere(this.props.focus);
+      const maybeNode = this.props.searchNodeEverywhere(this.props.focus);
+      if (!maybeNode) {
+        return;
+      }
+      const { node: n } = maybeNode;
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you!

PR #644, opened 2019-02-04
Fixing wrong state management in video plugin

DouglasGabr:
Problems fixed:
- Clicking outside of the `BottomToolbar` now calls the `TextField.onBlur` event. It was not calling because of the `focused` prop, it was unmounting the `BottomToolbar` component, therefor not calling `TextField.onBlur`;
- Save the `this.state.src` in a variable to send it in `this.props.onChange`. It was sending always `undefined`.


fernandobandeira:
Also, the last release (0.6.1) wasn't published to the npm, so maybe there's something broken with the deployments.

aeneasr:
Thanks!

PR #651, opened 2019-02-28
fix: name is not defined in Cell/Layout

tranvansang:
I wonder Why this doesn't break the whole package?


davidfurlong:
good question

PR #659, opened 2019-03-18
Add 'onRemoveCell' callback for plugins

agustin107:


davidfurlong:
After reviewing this I think it might be better to add the `onRemoveCell` prop and type to the `Decorated` component of Trash rather than `Provider`  as it has no business in the provider

```
const Trash: React.SFC<ProviderProps> = props => (
  <Provider {...props}>
    <Decorated {...props} />
  </Provider>
);
```

```
const Trash: React.SFC<ProviderProps & OwnProps> = ({ onRemoveCell, ...providerProps }) => (
  <Provider {...providerProps}>
    <Decorated onRemoveCell={onRemoveCell} {...providerProps} />
  </Provider>
);
```

macrozone:
@agustin107 is this still needed? I think we could merge this if you adjust to @davidfurlong s feedback (and rebase with master first)

macrozone:
I close this, feel free to reopen!

PR #670, opened 2019-04-22
Add Text tab to Background plugin (edit forecolor)

isvaljek:
I needed the option to set the text/foreground color for the Background component.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ coverage/
 lib/
 dist/
 examples/build
+.history
--- a/examples/src/index.tsx
+++ b/examples/src/index.tsx
@@ -102,7 +102,7 @@ const plugins: Plugins = {
     background({
       defaultPlugin: slate(),
       imageUpload: fakeImageUploadService('/images/sea-bg.jpg'),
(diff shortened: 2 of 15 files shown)


davidfurlong:
Thanks for taking the time to contribute this PR. I'm wondering whether adding text color to the Text plugin is a better pattern as the background plugin currently inserts a text node within it when created (which also may not be the most intuitive thing to begin with). The automatic inversion of text color when inside a background (which is the current behaviour) clearly doesn't work when the background is light colored, so there is certainly a need to be able to control either the lightness of the text or its color entirely.

Since there have been some reviewer/maintainer changes to this repo recently I'd be curious to hear their thoughts on this.

fernandobandeira:
I think we can add the color option to the slate plugin directly as you've said, it'll solve this problem and also add more feature to the editor as a whole.

macrozone:
should be now much easier to define to create a color control for slate. In fact, the example in the current beta defines a custom color plugin:

see https://github.com/react-page/react-page/blob/feat/uniforms-in-slate/examples/src/customSlatePlugin.tsx

PR #695, opened 2019-07-19
Feat/configurable slate plugins

macrozone:
## Proposed changes

Slate customization was mentioned in the docs, but it was not implemented in a way that devs can easily select which plugins to include or to customize the rendering behaviour.

This PR does the necessary refactoring, enhance the plugins to accept a custom rendering function and updates the documentation.

## Types of changes

## Checklist

- [x] I have added necessary documentation (if appropriate)

## Further comments

The refactoring is not perfect and the plugin codes still have a lot of repetitions. My goal was to make it work for developers that use the library. It can be further refactored and simplificated in a later iteration.

Biggest problem was that the `hooks` (not actual react hooks, but just some functions that get attached to the plugin for serialization / deserialization) need to know the plugins that are used, but the default plugins where just statically initialized and imported, so it was impossible to provide them runtime configuration.

Edit:

additionally, i changed the `StaticComponent` of the plugin to render real react, not just doing some `dangerouslySetInnerHtml`, which would lead to unexpected results when you render some more advanced components. Rendering pure html can be done using ReactDOMServer


macrozone:
tests run locally, not sure what's wrong with travis

PeterKottas:
The first pointer, pls get prettier formater when working with this lib. Cause I am getting a lot of false diffs cause by your formater.

macrozone:
> The first pointer, pls get prettier formater when working with this lib. Cause I am getting a lot of false diffs cause by your formater.

i got prettier, the initial commit of these files might be without prettier

packages/plugins/content/slate/src/__tests__/serialize.test.ts
```diff
@@ -243,14 +247,18 @@ describe('serialize to html', () => {
   ].forEach((c, k) => {
     describe(`test case ${k}`, () => {
       it('should serialize properly', () => {
-        // tslint:disable-next-line:no-any
-        expect(html.serialize(Value.fromJSON(c.i as any))).toEqual(c.o);
+        expect(
+          // tslint:disable-next-line:no-any
+          serializationFunctions.slateToHtml(Value.fromJSON(c.i as any))
+        ).toEqual(c.o);
```
macrozone: @PeterKottas you are referring to wrong diffs like this, right?

looks indeed like the previous version was without prettier or with a different configuration.

Or the line length is different

PeterKottas:
Damn, there are so many formating changes that it's really hard for me to figure out what exactly is going on. Overall I must say it's a very nice and big piece of work. I do think the whole plugin will definitely require some more refactoring in the future. Right now, it's kind of crazy complex for what it is. 

That being said, let me see if I get one thing right.

In your comment, https://github.com/react-page/react-page/issues/580#issuecomment-513481243, you mentioned the problem you saw was that additional libs are bundled without a real need for them. However, the current implementation is not taking care of that (as far as I can tell).
if you do

```
import defaultPlugins from './plugins/defaultPlugins';
import * as slatePlugins from './plugins/index';
```

Inside index, and then you import index and bundle it, it will end up there anyway. That means it doesn't matter if it's imported in hooks or index. 

Moving on, it was actually possible (unless I am missing something and please correct me if I do) to replace plugins before. This work is definitely making it easier by introducing the 

```
packages/plugins/content/slate/src/plugins/index.ts
```

But you could in fact import them one by directly from the files and provide them to the constructor of slate.

On top of that, if you import anything from packages/plugins/content/slate/src/plugins/index.ts, all plugins will actually end up in the bundle anyway. That's how webpack works. That why it's recommended to do
```
import throttle from 'lodash/throttle';
// instead of 
import { throttle } from 'lodash';
// because the latter would include the whole thing. The same thing is happening in our example.
```

The only way to make this work properly is as follows:
Leave the main index.ts, mostly for backward compatibility.

Then introduce a new file called createPlugin. That should expose a function that takes in config for the whole plugin (the config should be refactored because not the constructor of slate is a mess. it should accept one object called config and that's it.)
In this create plugin file, you should be very careful about providing default value for the config. Everything that you import here will end up in the bundle. Take a look at
https://github.com/react-page/react-page/blob/master/packages/plugins/layout/background/src/createPlugin.tsx
and
https://github.com/react-page/react-page/blob/master/packages/plugins/layout/background/src/index.tsx
see how createPlugin don't provide the 
```
import BackgroundDefaultControls from './Controls/BackgroundDefaultControls';
import BackgroundHtmlRenderer from './Renderer/BackgroundHtmlRenderer';
```
That's where the magic happens. That's how you can "create" your own plugin that will only bundle the stuff that is necessary. For instance here, the 
```
import BackgroundDefaultControls from './Controls/BackgroundDefaultControls';
```
includes all imports from material-ui. That means you can provide your own BacgkroundControls by implementing the interface and get rid of material-ui totally. 

* side note: 
Notice the
```
import Icon from '@material-ui/icons/CropLandscape';
```
? You might think that sucks because it goes against what I said. Yes, you would have to install ```@material-ui/icons``` but this will only bundle one icon, not the whole lib. So that's ok.

Well ... I assume this is confusing as fuck. It's a complex topic but hopefully, this will shed some light into this. Hit me with any questions.

Fair warning though ... slate is definitely by far the most complex piece of work to refactor this way. The value is quite high though imho. The work in the background plugin should be a neat starting point. GL

macrozone:
> Damn, there are so many formating changes that it's really hard for me to figure out what exactly is going on. Overall I must say it's a very nice and big piece of work. I do think the whole plugin will definitely require some more refactoring in the future. Right now, it's kind of crazy complex for what it is.

i don't see that much formatting issues, which one do you mean? most changes are real changes.

> That being said, let me see if I get one thing right.
> 
> In your comment, [#580 (comment)](https://github.com/react-page/react-page/issues/580#issuecomment-513481243), you mentioned the problem you saw was that additional libs are bundled without a real need for them. However, the current implementation is not taking care of that (as far as I can tell).
> if you do
> 
> ```
> import defaultPlugins from './plugins/defaultPlugins';
> import * as slatePlugins from './plugins/index';
> ```
> 
> Inside index, and then you import index and bundle it, it will end up there anyway. That means it doesn't matter if it's imported in hooks or index.

it does not take care of the import problem at all, it was just to enable plugins to take configuration, which was impossible before, that should be tackled separatly

> Moving on, it was actually possible (unless I am missing something and please correct me if I do) to replace plugins before. This work is definitely making it easier by introducing the
> 
> ```
> packages/plugins/content/slate/src/plugins/index.ts
> ```
> 
> But you could in fact import them one by directly from the files and provide them to the constructor of slate.

problem was `defaultPlugins` wich was statically imported by the serialization functions. So even if you provided different plugins, the serialization would use them instead of your plugins.

> On top of that, if you import anything from packages/plugins/content/slate/src/plugins/index.ts, all plugins will actually end up in the bundle anyway. That's how webpack works. That why it's recommended to do
> 
> ```
> import throttle from 'lodash/throttle';
> // instead of 
> import { throttle } from 'lodash';
> // because the latter would include the whole thing. The same thing is happening in our example.
> ```

i understand that. This PR does not change anything in the semantics of what is beeing in the bundled. One tiny step would be to not import all defaultPlugins. This would be a breaking change, because `slate()` without argument will use all defaultPlugins. So we would need a `slateWithDefaults()` and `slate(myplugins)`

Having said that: the single slate plugins themself are not very big, just some of them, like the link plugin will pull in material-ui. 

would be awesome if it would just import the editor-components lazily when needed, i.e. when the editor is not in readOnly mode. But i am not sure if webpack supports `import()` functions inside of node_modules, maybe i give it a try soon

If we would manage to import lazily, we could just keep the architecture as-is to reduce bundle size and still be able to provide the full default editor UI!

> The only way to make this work properly is as follows:
> Leave the main index.ts, mostly for backward compatibility.
> 
> Then introduce a new file called createPlugin. That should expose a function that takes in config for the whole plugin (the config should be refactored because now the constructor of slate is a mess. it should accept one object called config and that's it.)
> In this createPlugin file, you should be very careful about providing default value for the config. Everything that you import here will end up in the bundle. Take a look at
> https://github.com/react-page/react-page/blob/master/packages/plugins/layout/background/src/createPlugin.tsx
> and
> https://github.com/react-page/react-page/blob/master/packages/plugins/layout/background/src/index.tsx
> see how createPlugin don't provide the
> 
> ```
> import BackgroundDefaultControls from './Controls/BackgroundDefaultControls';
> import BackgroundHtmlRenderer from './Renderer/BackgroundHtmlRenderer';
> ```
> 
> That's where the magic happens. That's how you can "create" your own plugin that will only bundle the stuff that is necessary. For instance here, the
> 
> ```
> import BackgroundDefaultControls from './Controls/BackgroundDefaultControls';
> ```
> 
> includes all imports from material-ui. That means you can provide your own BacgkroundControls by implementing the interface and get rid of material-ui totally.

But that would also mean that initial setup would be more complex for users, right? 

> 
> * side note:
>   Notice the
> 
> ```
> import Icon from '@material-ui/icons/CropLandscape';
> ```
> 
> ? You might think that sucks because it goes against what I said. Yes, you would have to install `@material-ui/icons` but this will only bundle one icon, not the whole lib. So that's ok.

yes, i thought that as well ;-) still not so good that it is imported anyway when not using the editor part.

> Well ... I assume this is confusing as fuck. It's a complex topic but hopefully, this will shed some light into this. Hit me with any questions.

thx! i also have some experience with lazy loading and bundle resizing. The problem is more the monolytic architecture of react-page at the moment.
> 
> Fair warning though ... slate is definitely by far the most complex piece of work to refactor this way. The value is quite high though imho. The work in the background plugin should be a neat starting point. GL

i will take a look at that as well. But having said that, refactoring the plugin to reduce bundle size is not the goal of this PR

packages/plugins/content/slate/src/plugins/Plugin.ts
```diff
@@ -49,12 +67,12 @@ export default class Plugin {
    */
   public schema: {
     // tslint:disable-next-line:no-any
-    document?: { [key: string]: any };
+    document?: { [key: string]: any }
```
macrozone: ok these are the more annyoing formatting changes. Strange that the semicolons got lost here. i try to figure out why.

macrozone: ok i added a .prettierrc file to force semicolons, not sure why in some files it kept semicolons and in others not.

PeterKottas:
Afk now, I just glanced over your comments and all seems good to me. Will give you a proper response in a few hours. Cheers

PeterKottas:
> problem was defaultPlugins wich was statically imported by the serialization functions. So even if you provided different plugins, the serialization would use them instead of your plugins.

You are totally right about the issue with hooks! My bad, this makes sense now. 

> it does not take care of the import problem at all, it was just to enable plugins to take configuration, which was impossible before, that should be tackled separately

Cool, I thought so about the "import hell", it was just that comment on the other issue that threw me away. If this is not about making the bundle smaller, looks fair enough to me. Would need some extra testing though just to make sure.

> But that would also mean that initial setup would be more complex for users, right?

Actually, not at all, the experience for most users is exactly the same, you just import from the index and you're good to go. The thing this allows advanced users to do though is replacing the editor/rendering pipeline by totaly throwing away all code (and imports) we provide by default.

> i will take a look at that as well. But having said that, refactoring the plugin to reduce bundle size is not the goal of this PR

Glad to hear that, would be awesome to have this work in there. See what you can do, cheers!

fernandobandeira:
This is looking good, I'll give it some testing this week, my only concern is that the failing test on travis is signalling a possible breaking change, I'll take a closer look there to confirm that's not the case, but anyways great work here, ty for getting all of this done.

macrozone:
> This is looking good, I'll give it some testing this week, my only concern is that the failing test on travis is signalling a possible breaking change, I'll take a closer look there to confirm that's not the case, but anyways great work here, ty for getting all of this done.

locally the test works, but I'll check

macrozone:
can someone review and maybe help with the failing test?

macrozone:
this PR is now superseed by https://github.com/react-page/react-page/pull/700 as it includes this as well.

fernandobandeira:
I think we can close this in favour of #700, we should be able to also merge #699 #701 and #704 and release a new minor version with the breaking changes.

macrozone:
@fernandobandeira cool, i will try to do the release. I just need to push a tag on master, right?

Issue #698, opened 2019-08-07
lazy loading to reduce bundle size


macrozone:
this is an alternative approach to https://github.com/react-page/react-page/issues/580, at least when it comes to bundle size.

I think it's good to have a default UI for the editor. Material-UI is solid for that. 

The Problem at the moment is, is that if you want to use react-page also to display the content statically (without editing functionality), it still bundles material-ui, controls, etc.

https://github.com/react-page/react-page/issues/580 suggests (as far as i understand), that we should change the plugin system in a way, that the developer imports both the Control and the static Renderer and a compose function that creates the plugins out of the two. Or you import only the static Renderer and an empty Control if you want to only use the html renderer. While this works and additionaly provides the ability to swap Controls and renderes, it is more complex for devs to set it up. Also if you want to have both Editor and StaticRenderer in your app, you still have to deal with code spliting and have to thing how you describe your plugin configuration in a DRY way. Also the current plugin structure would need more refactoring as it is a bit verbose and not super clean atm.

I therefore play around with a different approach of code-splitting with the soon-to-be-standard `import()` function. Here are my findings so far:

- using `import()` and `React.lazy` work sort-of out of the box in the current setup, but projects will still bundle the whole thing, because tsc compiles `import()` to `require()``
- solution is to ` "target": "es6"` with ` "module": "esNext"`, but this will output plain es6 to lib. This can only be consumed if your project-runtime supports transpiling of node_modules. 
- In `nextjs` this can be solved with https://github.com/martpie/next-transpile-modules, `meteor` also supports transpiling node_modules with a symlink-trick at the time of writing
- i did an experiment with the image-plugin and can confirm that it successfully code splits the controls! I will send a PR to show it and will do further experiments

So what i would suggest is:

- every Control or basically everything that is renderer conditionally (typically when readOnly is false) should be a `React.lazy(() => import("./path/to/control"))`-component 
- we need to add `<Suspense />` gate at some point or use react-loadable as an alternative
- To keep backward compatibility, we would keep the default es5 output in /lib for the time beeing, but additionally provide a /es output directory or publish additional packages with -es suffix. We could later swap the default, because es-npm packages will become the default in the future
- projects that would want to profit from code splitting would need to import these es-packages and would need to configure transpiling of these packages (e.g. with https://github.com/martpie/next-transpile-modules)

macrozone:
ok, its even easier now!

- as written above, compile to target: es6 into folder lib-es
[...]

PR #700, opened 2019-08-07
Code splitting & bundle size reduction

macrozone:
this implements code split as described here https://github.com/react-page/react-page/issues/698

its based on https://github.com/react-page/react-page/pull/699

I also included https://github.com/react-page/react-page/pull/695 because i wanted to also lazy load slate

I will rebase once these PRS are approved and merged

**Further refactorings i did:**

- i slightly changed to structure of the plugins: `Controls` now contains only the controls (as the name suggests), it used to also call the `Renderer`, but that was not so clean. Now, the Renderer is called in the plugins `Component` and conditionally lazy-loads the `Controls`, if readOnly is false.
- i exported some types into their own files, because some types were imported across packages and i was not sure, if this would also lead to that whole file beeing bundled
- i added a bundle analyzer in the examples to see the different webpack bundles.
- **I removed support for StaticComponent**, it was only used for slate and it was confusing with the introduction of `Renderer`. This might be breaking for external plugins, but i think its good to simplify the api. Having said that, we further refactor this, so that a plugin always accepts a Renderer (or name it again StaticComponent) and a Control. When the editor is in readOnly, every plugin renders its Renderer, when not, it renders Renderer AND its Control. This logic is now still in the plugins, but could be abstracted away. Currently one Plugin - slate again - slightly differs from this logic: it does not render Renderer, when in edit mode, because it's Control component also includes the whole slate editor, but that could be fixed.

**What's next:**

- i hope this pr can be merged. I know, it's a lot, but the changes are more simple than it looks like
- publish a test version: i would like to publish a test version so that people can try this with in different environents


macrozone:
hooray, tests work now! (there is something wrong on master with the @types versions and some other typescript problems, this has been fixed here)

macrozone:
sorry for the messy comit history, i will do a better rebase

PR #660, opened 2019-03-18
Update @material-ui/lab version

agustin107:
Fix #656

Is an issue of @material-ui. Link: mui-org/material-ui#14686

PR #701, opened 2019-08-07
updated MaterialUI to 4.3 and React to 16.8.6

### LINKED ISSUE #693: Not working with material ui 4
**Describe the bug**
getting error: TypeError: styles_1.createGenerateClassName is not a function

ory-editor-ui/lib/ThemeProvider/index.js

  72 | var index_1 = __importDefault(require("./DarkTheme/index"));
  73 | 
  74 | exports.darkTheme = index_1.default;
 > 75 | var generateClassName = styles_1.createGenerateClassName({
  76 |   productionPrefix: 'ory'
  77 | });
  78 | exports.themeOptions = {
import React from 'react'
import Editor, { Editable, createEmptyState } from 'ory-editor-core'
import image from 'ory-editor-plugins-image'
import 'ory-editor-plugins-image/lib/index.css'

const editable = createEmptyState()

const editor = new Editor({
  editables: [editable],
  plugins: {
    content: [image]
  },
})

function BaseEditor() {
  return (
    <Editable editor={editor} />
  )
}

export default BaseEditor

**To Reproduce**
1. Update material ui package to 4.x
2. Start application
3. Look at console output

**Expected behavior**
Works with material ui 4

**Desktop (please complete the following information):**
 - OS: every os
 - Browser every browser
 - Version every version

Fivedark:
## Proposed changes

Updated dependencies
@material-ui/core to 4.3.0
@material-ui/icons to 4.2.1
@material-ui/styles to 4.3.0
react and react-dom to 16.8.6

Fixed issues caused by update mentioned in https://material-ui.com/de/guides/migration-v3/:
- MaterialUI Slider moved from lab to core for example.
- removed useNextVariants
- createGenerateClassName moved from core/styles to styles

Updated also all peerDependencies of all packages.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
   (To avoid dependency problems caused of new React and MaterialUI versions.)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (nothing to test)
- [ ] I have added necessary documentation (not appropriate)

## Closing issues

closes #693 
maybe also #660


andreastoermer:
Strange, ich nutze auch Node 10 und habe es gerade lokal getestet und da läuft es durch.

macrozone:
> Strange, ich nutze auch Node 10 und habe es gerade lokal getestet und da läuft es durch.

weird error in the CI... maybe some react version missmatch? maybe check the package-lock and yarn.lock s

Liebe Gruess aus der Schweiz ;-)

andreastoermer:
Oh sorry i missed to write in english :D

I mean, that it works local with Node 10.

macrozone:
maybe the problem is because the project mixes package-lock.json and yarn.lock

Depending on what you use that could lead to missmatches, because you might forget to update the other

packages/core/package.json
```diff
@@ -18,8 +18,8 @@
     "clean": "rimraf \"lib/*\""
   },
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
```
macrozone: this will require that consumers update to react 16.8.0 at least. Is there a reason why we force this version? does material 4 require react 16.8.0 or do we use hooks ?

examples/package.json
```diff
@@ -15,10 +15,11 @@
     "@react-page/plugins-video": "0.0.0",
     "@react-page/renderer": "0.0.0",
     "@react-page/ui": "0.0.0",
-    "react": "^16.6.3",
-    "react-dom": "^16.6.3",
-    "@material-ui/core": "3.6.1",
-    "@material-ui/icons": "3.0.1",
+    "react": "link:../node_modules/react",
+    "react-dom": "link:../node_modules/react-dom",
```
macrozone: so this will use the version from the parent right? 

In the parent, react is just a dev-dependency, while for the example it is a prod dependency. I would not mix that. This could break if the example gets autodeployed somewhere using just production node_modules.

examples/package.json
```diff
@@ -15,10 +15,11 @@
...
+    "@material-ui/core": "4.3.0",
+    "@material-ui/icons": "4.2.1",
+    "@material-ui/styles": "4.3.0",
```
macrozone: since we have a package-lock.json, we could relax the version here (with ^)

packages/core/package.json
macrozone: aaah, material-ui v4 requires react 16.8.0... so this is absolutly correct!

examples/package.json
Fivedark: I didn't get the examples to work without this linking, caused of two differenct react instances. This was the only way I got it to work on my local machine. If you have any other ideas, how to prevent to get two react instances, let me know.

examples/package.json
Fivedark: I will try to solve the pipeline issue and will allow also higher versions for the material ui dependecies today.

macrozone: maybe because of a version missmatch in package-lock.json or yarn.lock?

examples/package.json
macrozone: maybe because of a version missmatch in package-lock.json or yarn.lock?

Fivedark: I don't get `npm install` to run through. So I can't update the package-lock.json of examples. Maybe because I'm running a windows system. With yarn is everything running fine. I also was able to remove the react links, don't know why this wasn't possible last week. 

I'm hoping that one of my colleges can run the `npm install` on his machine and commit the updated lock file to the fork.

macrozone:
@Fivedark can you try to include https://github.com/react-page/react-page/pull/707 which will remove the package-lock.json files and do a `yarn install` in the root directory and check whether you have changes in the yarn.lock files?

Fivedark:
I think you missed one `package-lock.json` in the renderer package or was that intention?

macrozone:
> I think you missed one `package-lock.json` in the renderer package or was that intention?
> I removed it for now.

aah yes, i forgot it

macrozone:
@fivedark can you resolve the conflicts? thx!

Fivedark:
After resolving the conflicts, I'm getting again the follwing error when running the example:
```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
```
Last time I was able to solve this by linking the example react to the parent react:
```
    "react": "link:../node_modules/react",
    "react-dom": "link:../node_modules/react-dom",
```
But this isn't working anymore and there are also some concerns about this handling. (see above)

I'm sorry but I can't solve this issue now and I'm on holiday for 5 days starting tomorrow. So I pushed my current state to a new branch:
https://github.com/teamshirtsde/react-page/tree/rebased-material-ui-4.3

Feel free to solve the example issue to get the new react-page version online.

macrozone:
> After resolving the conflicts, I'm getting again the follwing error when running the example:
> 
> ```
> Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
> ```
> 
> Last time I was able to solve this by linking the example react to the parent react:
> 
> ```
>     "react": "link:../node_modules/react",
>     "react-dom": "link:../node_modules/react-dom",
> ```
> 
> But this isn't working anymore and there are also some concerns about this handling. (see above)
> 
> I'm sorry but I can't solve this issue now and I'm on holiday for 5 days starting tomorrow. So I pushed my current state to a new branch:
> https://github.com/teamshirtsde/react-page/tree/rebased-material-ui-4.3
> 
> Feel free to solve the example issue to get the new react-page version online.

will do! but if you have some spare time:

- check all yarn.lock files if the react-versions are exactly the same
- maybe remove all node_modules folders and do a `yarn install` in the root folder

Fivedark:
I already checked both of your suggestions. I think the problem is not that there are different versions of react but different instances or something like this.

rboixaderg:
> I already checked both of your suggestions. I think the problem is not that there are different versions of react but different instances or something like this.

Hi!
If you move react dependences to peerDependences, in the examples folder, the examples work done!

I think that the problem is that the packages are using the parent react version and the examples are using his own react version and this causes the error described above. 

Solution:

examples/package.json
`"peerDependencies": {
    "react": "^16.8.6",
    "react-dom": "^16.8.6"
  },`

In addition you can move material-ui in peerDependences too.

Fivedark:
It seem to work with peer dependencies, thanks @rboixaderg.

macrozone:
is it possible to make it backwards compatible with material-ui <4 ? i use react-page alongside react-admin, which is still on version 1 (or 3, not sure) (But new major version in alpha stage with material-ui 4 is in the making).

Or we could change material-ui do be in dependencies, instead of peer-dependencies, as we lazy-load it now, the user should not bother about it. we just use it to display the ui

Fivedark:
I think this could be difficult, because of the breaking changes between Material-UI 3 and 4 and the code changes on react-page to handline these.

fernandobandeira:
@macrozone I think we can move them to dependencies, makes more sense to me.

Fivedark:
Ok, I moved the material-ui everywhere from peer-dependencies to dependencies.

fernandobandeira:
Merged it, the only thing is that I we're using npm instead of yarn, I'll update the locks and remove the yarn ones, and add a gitignore so we prevent this.

macrozone:
@fernandobandeira should we just release 0.9.0? instead of -beta? we are still on 0.x.x, so we can break

andreastoermer:
@macrozone maybe it's time for a stable 1.0.0 release ;) Or are there motives, while we are still at 0.x.x?

macrozone:
@andreastoermer well, you are allowed to break the api more in 0.x releases ;-)

I would like to go with 1.x.x, but we might need better semantic release first. Lerna supports semantic releases, but its not used in this project yet

Fivedark:
And I think the documentation should be up to date for a stable 1.0.0.

PR #702, opened 2019-08-08
WIP: Lazy load parse5 for bundle size reduction

macrozone:
this is the next step in reducing bundle size. (based on previus pr https://github.com/react-page/react-page/pull/700)

now parse5 in slate plugin is loaded async when it's needed (it is needed when pasting html to slate or if you serialize from data that has `importFromHtml` properties).
This reduces the bundle size about 500kb unzipped and 110kb gzipped, which is huge if you think that an acceptable bundle size should be below 500kb gzipped

all plugin's unserialize functions now can be async and therefore return a promise (but they don't have to).

i did not make the same change to serialize, because i wanted to make this work first.


macrozone:
ok, the tests revealed a common problem: now the rendering is now somewhat async and react does not support async rendering properly (on SSR), so we need to find another solution.... what i now try is to just include parse5 when on server (or replace it with jsdom should it produce different results than the browsers DOMParser)

PR #704, opened 2019-08-09
Remove parse5

macrozone:
remove parse5 and replace it with jsdom on server and in tests. This reduces the bundle size by around 200kb.

also parse5 is not supported anymore by slate officially, altough it did not seem to have caused problems.

based on my previous refactor https://github.com/react-page/react-page/pull/700

we should merge that and then i will do a rebase here


macrozone:
hooray, tests still work ;-)

PR #712, opened 2019-08-19
feat: allow to define whether link should open in new window

macrozone:
slate link plugin now allow to define whether links should open in a new window

I added it while doing a refactor of the link plugin. I later want to allow for more complex plugins, like page links which store a pageId instead of a href and generate the href on render. The slate plugins are still a bit verbose and therefore hard to extend or customize. The goal is to have a similar flexibility like  with `@react-page/create-plugin-materialui`

## Proposed changes

slate link plugin has a checkbox which defines whether link opens in new window (`target="_blank"`)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Its *a little breaking*, but only if you override `getComponent` of the link plugin, which was a brand-new feature..

## Checklist

- [  I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/packages/plugins/content/slate/src/default/settings.ts
+++ b/packages/plugins/content/slate/src/default/settings.ts
@@ -10,6 +10,7 @@ export const defaultTranslations = {
     createLink: 'Create a link',
     linkTitlePlaceholder: 'Link title',
     linkHrefPlaceholder: 'http://example.com/my/link.html',
+    linkOpenInNewWindowLabel: 'Open in new window',
   },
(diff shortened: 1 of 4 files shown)


fernandobandeira:
>Its a little breaking, but only if you override getComponent of the link plugin, which was a brand-new feature..

I think it's fine, we could also wait to merge the material update in the other PR and tag this in a minor version then, but it should be alright to send as a patch release.

Issue #713, opened 2019-08-19
update slate


macrozone:
**Is your feature request related to a problem? Please describe.**

i sometimes get an error in slate, which seems to be fixed in most recent versions (have to look up the issue)

**Describe the solution you'd like**

update slate to the latest version

**Describe alternatives you've considered**

none

**Additional context**

This is breaking and needs some work. In particular in our slate plugins. See https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/Changelog.md#0220--may-8-2019

Issue #498, opened 2017-12-15
HTMLRenderer can't serialize type paragraph


miguel-orange:
When I copy and paste plain text into `Editable`, `HTMLRenderer` gives an error `Uncaught (in promise) Error: No serializer defined for node of type "paragraph".` I'm using the same plugins and when I refresh editable it has no issue rendering this.
There is only one difference between the two objects. The pasted text version has `"type": "paragraph"` vs `"type": "PARAGRAPH/PARAGRAPH"`.

Typed
`{
    "id": "",
    "cells": [
      {
        "id": "",
        "inline": null,
        "size": 12,
        "content": {
          "plugin": {
            "name": "ory/editor/core/content/slate",
            "version": "0.0.1"
          },
          "state": {
            "serialized": {
              "nodes": [
                {
                  "kind": "block",
                  "type": "PARAGRAPH/PARAGRAPH",
                  "nodes": [
                    {
                      "kind": "text",
                      "text": "The Northwest"
                    }
                  ]
                }
              ]
            }
          }
        }
      }
    ]
  }`
Pasted
`{
    "id": "",
    "cells": [
      {
        "id": "",
        "inline": null,
        "size": 12,
        "content": {
          "plugin": {
            "name": "ory/editor/core/content/slate",
            "version": "0.0.1"
          },
          "state": {
            "serialized": {
              "nodes": [
                {
                  "kind": "block",
                  "type": "paragraph",
                  "nodes": [
                    {
                      "kind": "text",
                      "text": "The Northwest"
                    }
                  ]
                }
              ]
            }
          }
        }
      }
    ]
  }`

developer-rakeshpaul:
@miguel-orange I also had the same problem. Where you able to solve this?. I wrote the below temporary hack to handle this. Just replaced the paragraph type with PARAGRAPH/PARAGRAPH before I save the content
```

export function replaceParagraphType(content) {
  const jsonString = JSON.stringify(content);
  const search = '"type":"paragraph"';
  const replacement = '"type":"PARAGRAPH/PARAGRAPH"';
  const newContent = jsonString.replace(new RegExp(search, 'g'), replacement);
  return newContent;
}

```

arekkas:
Oh that might be some issue with the copy/paste handler and specifically the HTML deserializer, we have to upgrade to the new Slate version anyways soon, maybe that will resolve those issues.

miguel-orange:
@developer-rakeshpaul, I handled it exactly how you did.

[...]

PR #716, opened 2019-08-20
Refactor slate

macrozone:
# Summarize:

- upgrades to latest slate
- adds new factories to create slate plugins
- allow to customize built-in slate plugins more easily

fixes https://github.com/react-page/react-page/issues/713

**BREAKING CHANGE:**

slate has been updated to latest version and contains some breaking changes. So if you use custom slate plugins, these might need some changes.

content however will be migrated thanks to the awesome migration system of react-page 👍 

**BREAKING CHANGE:**

The slate plugin customization of version 0.8.x of react-page no longer is supported.

If you had

```
const slatePlugin = slate([

  new slatePlugins.HeadingsPlugin({
    allowedLevels: [1, 2, 3],
    getComponent: ({ type }) => {
      return MyCustomHeadingsComponent
    },
  }),
// ... other slate plugins
]);
```

you now have to use this:

```
const slatePlugin = slate([

  slatePlugins.headings.h1(),
  slatePlugins.headings.h2(),
  slatePlugins.headings.h3(
    // customization is now more flexible, see below
   def => ({
      Component: MyCustomHeading3
   })
  ),

// ... other slate plugins
]);
```

## customize slate plugins:

if you just use `slate()` all default plugins will be loaded as before, so you don't need to change anything

### select plugins

if you don't want to use all plugins, you can define the list of all slate plugins yourself:

```

import slate, { slatePlugins } from "@react-page/plugins-slate";

// this is the list of all default plugins, remove the ones you don't need
const slatePlugin = slate([
    slatePlugins.headings.h1(),
    slatePlugins.headings.h2(),
    slatePlugins.headings.h3(),
    slatePlugins.headings.h4(),
    slatePlugins.headings.h5(),
    slatePlugins.headings.h6(),
    slatePlugins.link(),
    slatePlugins.lists.ul(),
    slatePlugins.lists.ol(),
    slatePlugins.lists.indention(),
    slatePlugins.emphasize.em(),
    slatePlugins.emphasize.strong(),
    slatePlugins.emphasize.underline(),
    slatePlugins.alignment.left(),
    slatePlugins.alignment.center(),
    slatePlugins.alignment.right(),
    slatePlugins.alignment.justify(),
    slatePlugins.paragraph(),
    slatePlugins.blockquote(),
    slatePlugins.code.mark(),
    slatePlugins.code.block(),
]);
```

also this allow to re-arange the plugins order

### customizing existing plugins:

every of these built-in-plugins consume a customize function. E.g. if you want to change the component that is used to render the plugin:

```
 headings.h1(def => ({
    ...def,
    Component: props => (
        <MyCustomH1 style={{align: props.data.get("align")}}>{props.children}</MyCustomH1>
   )
  })),
```

typescript users should get nice types in the customize function. Also you can provide custom controls, see the built-in link plugin how this can be done. We will provide a more convenient way to add controls to your slate plugin in the near future.

### create your slate plugin from scratch:

There are some factory functions that let you create plugins easily:

`createComponentPlugin`: allow to create a plugin which has a component (most built-in plugins use this factory).
`createDataPlugin`: this plugin toggles data on the current block or inline element, but does not render a component. E.g. the alignment-plugin uses that factory
`createSimpleHtmlBlockPlugin`: a more convenient variant of `createComponentPlugin`. It renders a simple component with a html-tag. used by plugins like `headings` or `blockquote` 
`createMarkPlugin`: similar to `createSimpleHtmlBlockPlugin` but renders a "mark" with a hover button on selection

you can import these with:

```
import {pluginFactories} from '@react-page/plugins-slate'

```

### initial comment

this is a big one

Makes creating new slate plugins as easy as:

```

const link: SlatePluginDefinition<LinkData> = {
  DEFAULT_NODE,
  type: 'LINK/LINK',
  object: 'inline',
  icon: <LinkIcon />,
  addHoverButton: true,
  addToolbarButton: true,
  Controls,
  deserialize: {
    tagName: 'a',
    getData: el => ({
      href: el.getAttribute('href') || '',
      openInNewWindow: el.getAttribute('target') === '_blank',
    }),
  },
  Component: ({ children, data }) => {
    return (
      <a
        target={data.get('openInNewWindow') ? '_blank' : undefined}
        href={data.get('href')}
      >
        {children}
      </a>
    );
  },
};

export default createComponentPlugin(link);
```

It enables to create new plugins much easier. In particular, because you don't have to deal with much of slate's complexity.

I wanted to refactor it, because i need to have a pageLink component in my project, which will not store the href of the target page, but its database-id and its custom link component will create the href based on that. Now this should be much easier and straight-forward to do. i will also add late a schema-based Control for it. 

In the current state, the Link still has a somewhat complicated `Controls` Component (which displays inputs to insert/update a link) and i want to make that much simpler


fernandobandeira:
This looks way better than what we had, it should also fix #498 and some other copy/pasting issues too, good work 👍

Fivedark:
Since this nice improvments were merged, I'm having trouble with the slate plugin on Firefox (Windows). The Cursor jumps after some key strokes on the end of the last text chunk. So the first letter is on the right position and all following appear on another position.
Chrome is working fine. 

Could somebody please check this behavior? 

Just checkout master, install and start (with Firefox). Try typing inside the text.

macrozone:
hi @Fivedark, can confirm this problem also on firefox on mac

PR #718, opened 2019-08-27
schema based slate plugins and more...

macrozone:
Goal was to provide schema-based forms for slate-plugins and to unify the plugin system even more, but i stumbled upon some troubles with material-ui etc. so i ended up refactoring and upgrading more than i wanted... _so i am very sorry for the big Pull request and then non-atomic commits_... But we have now some neat features:

**feat** the slate plugin itself now takes a customize function instead of an array of plugins. This makes it easier to customize it, but it's a **BREAKING CHANGE**:

_Before:_ `slate([plugin1, plugin2, plugin3])`
_After:_ `slate(def => ({...def, plugins: [plugin1, plugin2, plugin3]})`

**feat**: slate-plugin factories now take a schema, same as plugins made with
@react-page/create-plugin-materialui

The can render a custom form based on a jsonSchema.

The default link plugin now uses such a factory:

```
import createComponentPlugin from '../../pluginFactories/createComponentPlugin';
import React from 'react';
import { lazyLoad } from '@react-page/core';
const LinkIcon = lazyLoad(() => import('@material-ui/icons/Link'));

export type LinkData = {
  href: string;
  openInNewWindow?: boolean;
};

export default createComponentPlugin<LinkData>({
  type: 'LINK/LINK',
  object: 'inline',
  icon: <LinkIcon />,
  addHoverButton: true,
  addToolbarButton: true,
  schema: {
    type: 'object',
    required: ['href'],
    properties: {
      href: {
        type: 'string',
      },
      openInNewWindow: {
        type: 'boolean',
      },
    },
  },
  deserialize: {
    tagName: 'a',
    getData: el => ({
      href: el.getAttribute('href') || '',
      openInNewWindow: el.getAttribute('target') === '_blank',
    }),
  },
  Component: ({ children, data }) => {
    return (
      <a
        target={data.get('openInNewWindow') ? '_blank' : undefined}
        href={data.get('href')}
      >
        {children}
      </a>
    );
  },
});
```

very neat, isn't it? You want to allow your editor to add custom page links that store a pageId instead of a href? no problem now!

**feat**: slate plugin with controls (e.g. the link above) now allows to update the data (e.g. the href) and can be removed in the dialog

**feat**: @react-page/create-plugin-materialui functions are now more
similar to slate plugin factories, so they also return a function that
takes a customize function. this is a BREAKING CHANGE if you used that feature.

**feat**: plugin controls now render in white instead of black. That's actually a workaround, because the ThemeProvider from material-ui behaved very funky with the dark theme... The common slate-buttons are still black, though. Personally, i think white is better in this case. Also the left toolbar with all plugins is also bright and not dark.

**feat**: `@react-page/ui` now exports a default ui with Toolbar, Trash and DisplayMode-Buttons, to make it easier to setup. The motivation was to reduce the amount of Providers because i thought that they would mess with material-ui

**BREAKING CHANGE**:  `<Trash />`, `<DisplayModeToggle />` and `<Toolbar /> ` are no longer wrapped in a `<Provider />`. Use the new default export instead:

**Before**:

```
import { Trash, DisplayModeToggle, Toolbar } from '@react-page/ui';
// ...
<div>	  
    <Trash editor={editor} />	
    <DisplayModeToggle editor={editor} />	
    <Toolbar editor={editor} />	
</div>,
````

new: 

```
import EditorUi from '@react-page/ui';
// ... 
  <EditorUi editor={editor} />
```
   
if you still want to have these elements separated, you can wrap them manually with a provider:

```
import { Provider, Trash, DisplayModeToggle, Toolbar } from '@react-page/ui';
// ...
<Provider editor={editor}>	  
    <Trash />	
    <DisplayModeToggle />	
    <Toolbar />	
</Provider>,
````

internally, these components use the contextApi with a hook to get the editor. This might become handy in future iterations.

**feat**: redux has been upgraded to latest version

**feat**: material-ui has been updated to 4.3.3 and its now only defined in `@react-page/ui` and others use it as peer dependency if they depend on `@react-page/ui`

**fix**: broken in firefox because of https://github.com/ianstormtaylor/slate/issues/2978

**fix**: preventing blur now works through portals. This is handy if your slate plugins contain a portal (e.g. material-ui select-field uses a portal)

**chore**: fix some cross imports from libs

**chore**: fix some typescript compiler problems

**chore**: example has been updated to include a custom slate plugins

**chore**: typescript has been upgraded to 3.5.1. I had some funky problems with it, but it should now work more stable...


macrozone:
@fernandobandeira I know it sucks to review a pr like this, but can you give it a try? i will make better PRs in the future, but i got carried away by this one ;-)

it seems to work properly and i will test it more.

Also mind the smaller breaking changes, which should be documented. Also the documentation is updated (minus the api docs)

macrozone:
by the way, something is wrong with the push. It pushes properly, but it says it failed... that's why the failing check

fernandobandeira:
Cool, I'll review it today, np about the size of the PR this happens in other communities frequently too I think this'll be a great addition to the library, I'll check why it's failing and commit a fix if it's needed. 👍

macrozone:
> Cool, I'll review it today, np about the size of the PR this happens in other communities frequently too I think this'll be a great addition to the library, I'll check why it's failing and commit a fix if it's needed. 👍

i tagged the branch with a prerelease, so that we can try it out more easily. This does a push to npm, which works, but it shows in travis as failed and this updates also the travis-ci/push check here to beeing failed...

Issue #562, opened 2018-08-17
Automatically open edit form after adding a plugin via drag n drop


PeterKottas:
I have one more PR in mind but wanted to run in past you @arekkas 

We think it would make sense to give users a chance to setup the plugin right after dropping it in move mode. In other words, you drag an image plugin, drop it in place, a dialog would open allowing you to upload/modify the image, change link and so on. Then you click OK and can add more things right away.

Right now, if adding many images, you have to switch to edit mode every time which is time consuming. Yes it gives you a placeholder but why not set it up right away.

This wouldn't be necessary for all plugins (with slate it makes no sense for example), but especially image, video, space, background would benefit from this.

What I suggest is creating a new property on PLUGIN api called **dropControls**. These would be rendered in a dialog in the core package after drop. OK and CANCEL button would come from the LIB. Standard stuff as onChange would of course be injected in props.

arekkas:
This can be solved differently. You could switch into edit mode programmatically and then focus the new cell. This will yield the same result and not require another dialog.

PeterKottas:
I actually kind of like the dialog idea. It seems the expected ux ui experience. Think about it, you create something, you are asked for details. You set up, click OK and the items is created. Now it sort of feels like this could be achieved the way you are mentioning. Basically conditionally rendering extra Buttons in the UI. But then again, while in this TEMP edit mode, you could click on something else. And that's deffo not good UX UI. 

If I didn't have to take the implementation complexity into consideration, an extra dialog seems like the best user experience. I'll think about this a bit more. Maybe do some POC on our end.

arekkas:
I really want to avoid having any type of UI components in the core. The core is there for managing state, transitions, and making sure that the edit controls (drag & drop, resize, focus) work properly. Messing with this concept will degrade the code quality and break the packages isolation.

Additionally, having another dialog which is configured differently from the "edit view" is, in my opinion, bad developer experience. There should be one definition of how editing works, not multiple. The same goes for the state - there should be a clear "edit" state and a "view" state. An in-between edgecase state increases complexity and muddies the definition of the modes.

Also, in my experience, I like to define the layout first, then add the content. So typically, I add the boxes where I want them to, then add the images. This is personal preference obviously, but I think not having to click cancel every time I drop a certain plugin is something I want.

Instead of adding another function to the plugin API, I think it would make more sense to make this configurable when creating the editor instance. Maybe something like having callbacks such as `onDrop` or whatever.

[...]

PR #722, opened 2019-08-30
feat: switch to editmode and show controls if you add new plugins

macrozone:
fixes https://github.com/react-page/react-page/issues/562

I decided to use redux-thunk for that, because i think it's the right approach in our case, where we have actions for all the ui staff in place. redux-thunk is very small library that allows to dispatch multiple other actions on one action. it also allows to do async stuff.

The thing that confused me, was how all these cell actions defined an array `ids` with random generated ids. In case of insert actions, it defined an array of 5 ids. When cells and rows got created, it used the ids from this "pool" of ids. The different positions in these array were used as if they had no semantics, so they werent used consistently.

I noticed that, because i neeed to know which id the element will have that will contain my inserted plugin. I though that would be always the first id, but that was not the case. So i decided to change the array to an object where i could give the ids some meaning. In particular, the element that will contain the plugin `item` is always `ids.item`. So in my thunk, i can now use this id to focus this element and show its controls. it needs some cleanup though and the purpose of this ids object is still confusing, but at least it works now.

Now, if you add a new plugin, it will switch to editMode and put the focus state on the added plugin. This will show the controls.

It does not actually "focus" an html element, it just set's editors' focus state on that plugin. Switching the browsers focus to some meaningful element would be cool though. E.g. in case of slate, it should set the cursor into the text. In case of a plugin with a form, it should focus the first input of that form


packages/ui/src/Toolbar/Draggable/helper/index.ts
```diff
@@ -32,6 +32,7 @@ export const source = {
 
   // tslint:disable-next-line:no-any
   endDrag(props: any, monitor: any) {
+    return;
     const item = monitor.getItem();
     if (monitor.didDrop()) {
       setTimeout(() => {
```
macrozone: need to decide what to do with that. i think  its actually not needed anymore.

packages/core/src/store.ts
```diff
@@ -48,6 +48,6 @@ export default (initialState: any, middleware: [] = []): Store<RootState> => {
   return createStore(
     rootReducer,
     initialState,
-    v(applyMiddleware(...middleware))
+    v(applyMiddleware(...middleware, thunk))
```
macrozone: we now need the thunk middleware. I think we need to mention that, if you pass your own store to it

fernandobandeira:
Focusing the html element would be great indeed, for now this is cool.
Love the ids object it'll help us guarantee that we're accessing the correct cell.
As for the thunk middleware, it should be fine, we just add a note, it's a common middleware too so it shouldn't be inconvenient to the users.

Issue #761, opened 2020-01-04
No longer able to provide react-dnd backend


PeterKottas:
Hey, @macrozone and happy new year! I was doing some tests on mobile and after some digging I realized that as part of upgrading react-dnd, you got rid of the react-dnd backend injection api. https://github.com/react-page/react-page/commit/ceaecc192f0751a32793ccb5fe387ce1bee5ad0a#diff-55646d75364ce42cdb84abec1a1531ccL122
I was curious if you could shed some light on this, I assume it was something that changed with the latest version, however, I don't think it should be just removed. This was for instance discussed here before, https://github.com/react-page/react-page/issues/424 but after this change, it no longer works. I think we should discuss how to bring this back, hope there's no major reason why it was removed.

macrozone:
@PeterKottas the goal was to be able to have independent editor instances, so that they can co-exist. Therefore every editor has its own context with redux and dnd-backend. This was not possible in earlier versions of react-dnd as it lead to this error: https://github.com/react-page/react-page/issues/424

Passing your own dnd-backend is no longer needed. But maybe you have a usecase for it?

PeterKottas:
Ok I see, that makes sense. However, the issue remains as folks (including me) might want to on some occasions replace the default backend with touch backend. https://react-dnd.github.io/react-dnd/docs/backends/touch . As it stands now, it doesn't work on mobile at all.

macrozone:
@PeterKottas i think we can either add a mobileSupport flag and provide a touch backend or we could allow to pass one. Because there are only these twos, it might be better to enable touch support ourself

PeterKottas:
Not a good idea to go with the flag. Imagine one day, somebody comes up with a new (better) touch backend. Or entirely new backend for some specific hardware (iot/wearables ... who knows). Better have an api to provide backend from outside with reasonable default (as it was before). Because otherwise, sometime along the way, somebody will have to deal with this again.

macrozone:
@PeterKottas i think that‘s a YAGNI ;) better they would file an issue that they need support for an editor on their wearable and we could find a proper solution to fullfill this request. I think the backend stuff of react-dnd is such an implementation detail that i would not want to expose this to the api surface of react-page. I think the same about redux. They should stay internal. 

On the other hand, just passing a a touch backend to the provider would also be easy, but i would avoid it

PeterKottas:
[...]

PR #762, opened 2020-01-06
feat: allow to pass react-dnd-backend

macrozone:
You can now  customize the drag n drop backend from react-dnd. See https://react-dnd.github.io/react-dnd/docs/overview#backends for more information

example 

```
import Editor from '@react-page/editor';

import TouchBackend from 'react-dnd-touch-backend-cjs'
// ...

<Editor dndBackend={TouchBackend} />

```

or if you use the older Editable / EditorUI pair:

```
import EditorUi from '@react-page/ui';
import Editable from '@react-page/core';

import TouchBackend from 'react-dnd-touch-backend-cjs'


// ...

<Editable dndBackend={TouchBackend} /> 

// ...
<EditorUi dndBackend={TouchBackend} />

```

fixes https://github.com/react-page/react-page/issues/761

@PeterKottas can you test it a little?

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -20,6 +20,7 @@
  *
  */
 
+import { BackendFactory } from 'dnd-core-cjs';
 import forEach from 'ramda/src/forEach';
 import { NativeTypes } from 'react-dnd-html5-backend-cjs';
 import { Middleware, Store } from 'redux';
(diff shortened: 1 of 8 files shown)


packages/core/src/index.ts
```diff
@@ -52,6 +53,7 @@ import createStore from './store';
 import { EditableType, NativeFactory } from './types/editable';
 import { RootState } from './types/state';
 
+export { BackendFactory as DndBackend };
```
PeterKottas: I think we should get rid of this and import directly from the package where needed.

macrozone: problem is that you need to install `react-dnd-core` in your application just for that type. It is not exportet from `react-dnd`. I want to avoid make it more complicated for users.

Issue #757, opened 2019-12-04
update slate to 0.50


macrozone:
slate 0.50 is out: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/Changelog.md

https://docs.slatejs.org/concepts/xx-migrating

it requires some refactoring, but should be an intesting thing to do. We would profit from reduction in bundle size and complexity and probably also better performance

PeterKottas:
Hey mate, this is a plugin that fully supports the latest slate, I didn't have time to contribute here as I need to move fast, however, it could be good for inspiration (even copy-paste).

macrozone:
@PeterKottas ok, i will take a look at it

PR #765, opened 2020-01-08
update to slate ^0.50.0

macrozone:
This one is a tough one. But nearly working now!  The hardest problem was the list-indention stuff... i thought i would be a decent programmer, but i felt stupid with that one...

It would not have been possible without typescript! Also thanks to @PeterKottas  for the inspiration, I hope your custom plugin will be obsolete soon ;-)

TODO:

- [x] fix undo / redo
- [x] fix list indention (still funky)
- [x] pasting works, but seems to remove headings? check if this happens on slate official example as well

Changes so far:

**feat**: update to latest slate 0.57.0 (obviously). This should greatly reduce the bundle size  _(no immutable.js!)_ and improve performance, altough at the time of writing unessesary rerenders may occure as they are not yet optimized. 

**feat**: switching between read-only and edit mode does no longer rerender all the slate elements

**feat**: the slate plugin has now a function `htmlToSlate` which returns slate state based on the given plugins

**BREAKING CHANGE**: list plugins have slightly changed, the definition is a bit different. This is only a breaking change if you have customized lists. typescript should make it obvious what to do, will update the docs as well

**BREAKING CHANGE**: slate plugins have now internally a different structure, so completly custom slate plugins will no longer work. However, if you use the provided factory functions or customize the default plugin, you are not affected.

fixes https://github.com/react-page/react-page/issues/757


PeterKottas:
You're welcome mate and GJ as always. Looking forward to seeing this released! I will check the code once the TODO is done.

macrozone:
@PeterKottas seems to work quite nice so far, feels much faster and also the bundle size has been reduced.

There will be a follow up pull request that fixes some minor bugs and the following:

- fix hotkeys (global hotkey handler was actually broken for some hotkey)
- fix undo/redo when there are more then one editors on the page (using @react-page/editor)

there will be a small breaking change concerning usage without @react-page/editor: you have to wrap all Editables and EditorUI with a new `<Provider />` from @react-page/core. This will make it more future proof for further improvements and refactorings. One next step will be to rewrite some code to use hooks and redux-hooks. This will drastically simplify the code base and reduce the need of type annotations (`connect()` and `mapDispatchToProps` is just horror compared to using hooks) 

I think this one is ready to merge i think, alltough there might be some minor flaws. Since i already moved on with more changes (i currently feel very comfortable in the code-base), I would like to do corrections and fixes in upcoming PRs to reduce friction.

PR #793, opened 2020-04-16
[Fix] Only render tooltip in active buttons

rboixaderg:
Create a ConditionalWrapper component to only render Tooltip if button is active. 

## Proposed changes

Create a new component, to avoid material ui messge in disactive slate buttons:

```
Material-UI: you are providing a disabled `button` child to the Tooltip component.
A disabled element does not fire events.
Tooltip needs to listen to the child element's events to display the title.

Add a simple wrapper element, such as a `span`.

```
Only render tooltip if button is active. 

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)


macrozone:
nice @rboixaderg !

just a small notice: can you check if your editor respects the prettier rules? it swaped single with double quotes.

rboixaderg:
Oh yes! 

Changed @macrozone

PR #795, opened 2020-04-22
Do not manipulate data copied from slate editor

nhkhanh:
## Proposed changes

We should keep data copied from slate editor as-is instead of manipulate it by text/html or plain/text.

## Types of changes

## Checklist

- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules


macrozone:
@nhkhanh makes sense! thank you!

Issue #800, opened 2020-05-17
BlurGate


PeterKottas:
**Describe the bug**
There is a new component that was introduced a little while ago called [Provider](https://github.com/react-page/react-page/tree/46a690f8a5b6849a6fadb23b0b1c66818bceb83e/packages/core/src/Provider). Inside this component is another component called [BlurGate.](https://github.com/react-page/react-page/blob/46a690f8a5b6849a6fadb23b0b1c66818bceb83e/packages/core/src/components/BlurGate.tsx)

This component is switching to edit mode whenever you click outside of Provider tree. There is currently no way to enable/disable this functionality from outside by external API and this is causing a problem on my end.

I will provide a fix for this in a little bit via PR

PeterKottas:
PR in https://github.com/react-page/react-page/pull/801, could you pls take a look @macrozone ?

Also, I am curious if we should maybe consider removing the BlurGate functionality altogether. It sounds more like a userland implementation than something that should live in the core. What do you think?

macrozone:
@PeterKottas the blur gate is really important for usability and should not be in userland. Its particularly important to work correctly, if you have multiple editors on the same page. 

Please report what is not working for you, instead of working around it. Maybe there is a bug or some glitch?

PeterKottas:
Hmm, pls help me understand. If I have 2 editors next to each other (for some reason), and I want to be in the move mode in one, and resize mode in the other, it would keep switching to edit mode on the other one every time I work with the first editor. Right? I don't think that's good. Unless I am missing some limitations regarding react-dnd maybe.

In my case, I have an interactive tutorial implemented using https://reactour.js.org/. I lead people to switch modes and explain how it works but unfortunately, with this, every time I click inside the tutorial modal, it switches the mode back to edit making it unusable. 

I can easily imagine other examples where this should be disabled. Imagine a sidebar implemented with a drawer component. That would once again cause possibly unwanted switching between modes. 

Also, why does it switch to edit mode? Wouldn't maybe preview mode make more sense? Generally speaking, if you leave editable component ut usually reverts to it's "readonly" state. Also maybe, should it restore the previous mode once click back into the blurgate? Once again, I am not seeing your use-case clearly so this whole paragraph might not make sense.

PeterKottas:
[...]

PR #801, opened 2020-05-17
Added props to disable BlurGate

PeterKottas:
Detailed in https://github.com/react-page/react-page/issues/800

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "lint:watch": "npm-run-all watch:ts -- -c \"npm-run-all lint\"",
     "prettier": " prettier-tslint fix \"packages/**/src/**/*.(ts|tsx)\"",
     "publish": "lerna publish",
-    "start": "TSC_WATCHFILE=UseFsEventsWithFallbackDynamicPolling npm-run-all -p build:watch start:examples",
+    "start": "cross-env TSC_WATCHFILE=UseFsEventsWithFallbackDynamicPolling npm-run-all -p build:watch start:examples",
     "start:examples": "cross-env NODE_ENV=development lerna run start --stream",
     "test": "cross-env NODE_ENV=test jest --coverage",
     "test:watch": "cross-env NODE_ENV=test jest --watch",
--- a/packages/core/src/Provider/index.tsx
+++ b/packages/core/src/Provider/index.tsx
@@ -5,10 +5,12 @@ import HTML5Backend from 'react-dnd-html5-backend';
 import BlurGate from '../components/BlurGate';
 import Editor from '../Editor';
 import { ReduxProvider } from '../reduxConnect';
+import { DisplayModes } from '../actions/display';
 interface ProviderProps {
   editor: Editor;
   dndBackend?: BackendFactory;
   blurGateDisabled?: boolean;
+  blurGateDefaultMode?: DisplayModes;
 }
 
 const EditorContext = React.createContext<Editor>(null);
@@ -20,11 +22,14 @@ const Provider: React.FC<ProviderProps> = ({
   children = [],
   dndBackend = HTML5Backend,
   blurGateDisabled = false,
+  blurGateDefaultMode,
 }) => (
   <DndProvider backend={dndBackend}>
     <ReduxProvider store={editor.store}>
       <EditorContext.Provider value={editor}>
-        <BlurGate disabled={blurGateDisabled}>{children}</BlurGate>
+        <BlurGate disabled={blurGateDisabled} defaultMode={blurGateDefaultMode}>
+          {children}
+        </BlurGate>
       </EditorContext.Provider>
     </ReduxProvider>
   </DndProvider>
--- a/packages/core/src/Provider/index.tsx
+++ b/packages/core/src/Provider/index.tsx
@@ -8,6 +8,7 @@ import { ReduxProvider } from '../reduxConnect';
 interface ProviderProps {
   editor: Editor;
   dndBackend?: BackendFactory;
+  blurGateDisabled?: boolean;
 }
 
 const EditorContext = React.createContext<Editor>(null);
@@ -18,11 +19,12 @@ const Provider: React.FC<ProviderProps> = ({
   editor,
   children = [],
   dndBackend = HTML5Backend,
+  blurGateDisabled = false,
 }) => (
   <DndProvider backend={dndBackend}>
     <ReduxProvider store={editor.store}>
       <EditorContext.Provider value={editor}>
-        <BlurGate>{children}</BlurGate>
+        <BlurGate disabled={blurGateDisabled}>{children}</BlurGate>
       </EditorContext.Provider>
     </ReduxProvider>
   </DndProvider>
--- a/packages/core/src/actions/cell/insert.ts
+++ b/packages/core/src/actions/cell/insert.ts
@@ -129,6 +129,14 @@ export const insertCellRightInline = insert(CELL_INSERT_INLINE_RIGHT);
 const newIds = ({ id, ...item }: Partial<Cell>) => {
   return {
     ...item,
+    content: item.content && {
+      plugin: item.content.plugin,
+      state: JSON.parse(JSON.stringify(item.content.state)),
+    },
+    layout: item.layout && {
+      plugin: item.layout.plugin,
+      state: JSON.parse(JSON.stringify(item.layout.state)),
+    },
     id: v4(),
     rows: item.rows
       ? item.rows.map(row => ({
--- a/packages/core/src/actions/display.ts
+++ b/packages/core/src/actions/display.ts
@@ -54,6 +54,19 @@ const setDisplayMode = (
 });
 
 /**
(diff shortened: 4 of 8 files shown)


PeterKottas:
Is the CI expected to fail? Cause this doesn't seem like something I touched.

macrozone:
> Is the CI expected to fail? Cause this doesn't seem like something I touched.

do you see the ci? the linting fails as you changed from single quotes to double quotes. Please check your editor, the project should contain some linting/prettier rules (or i hope so?)

PeterKottas:
Weird, it always worked out of the box. Maybe I have different defaults on this machine.

PeterKottas:
Mate, looking at https://github.com/react-page/react-page/blob/master/tslint.json#L67, I think it should be single. My machine agrees. So not sure why the CI is coughing up. Also, it's single in most other files.

macrozone:
> Mate, looking at https://github.com/react-page/react-page/blob/master/tslint.json#L67, I think it should be single. My machine agrees. So not sure why the CI is coughing up. Also, it's single in most other files.
> 
> > the linting fails as you changed from single quotes to double quotes
> 
> I think fact changed from double to single which should be correct according to tslint config.

yes it should be single, but you changed it to double quotes

package.json
```diff
@@ -61,7 +61,7 @@
     "lint:watch": "npm-run-all watch:ts -- -c \"npm-run-all lint\"",
     "prettier": " prettier-tslint fix \"packages/**/src/**/*.(ts|tsx)\"",
     "publish": "lerna publish",
-    "start": "TSC_WATCHFILE=UseFsEventsWithFallbackDynamicPolling npm-run-all -p build:watch start:examples",
+    "start": "cross-env TSC_WATCHFILE=UseFsEventsWithFallbackDynamicPolling npm-run-all -p build:watch start:examples",
```
PeterKottas, line 64: This is unrelated, just something that must have been added a while ago and it wasn't done properly (cross platform)

packages/core/src/actions/cell/insert.ts
```diff
@@ -129,6 +129,14 @@ export const insertCellRightInline = insert(CELL_INSERT_INLINE_RIGHT);
...
   return {
     ...item,
+    content: item.content && {
+      plugin: item.content.plugin,
+      state: JSON.parse(JSON.stringify(item.content.state)),
+    },
+    layout: item.layout && {
+      plugin: item.layout.plugin,
+      state: JSON.parse(JSON.stringify(item.layout.state)),
+    },
```
PeterKottas, line 139: This is what fixes the duplicate button problem mentioned in #802 . Sorry again about mixing PRs.

packages/core/src/Provider/index.tsx
```diff
@@ -5,9 +5,12 @@ import HTML5Backend from 'react-dnd-html5-backend';
 import BlurGate from '../components/BlurGate';
 import Editor from '../Editor';
 import { ReduxProvider } from '../reduxConnect';
+import { DisplayModes } from '../actions/display';
 interface ProviderProps {
   editor: Editor;
   dndBackend?: BackendFactory;
+  blurGateDisabled?: boolean;
+  blurGateDefaultMode?: DisplayModes;
```
macrozone, line 13: we should probably also add this in `Editor` from 'react-page/editor` as its now recommended to use just this simple component. 

but there i would name it `displayModeDefault` or `defaultDisplayMode`

do you still need the `blurGateDisabled` ? is there another issue with it?

package.json
macrozone: perfect

packages/core/src/actions/cell/insert.ts
macrozone: no worries :-) looks good. This would probably fail if some state happens to be circular (which should not be the case in my opinion)

packages/core/src/Provider/index.tsx
PeterKottas: Yp I'll add it there. Yeah, I need to disable it. Even the blurAllCells doesn't work for me. I don't want my bottom toolbar closing while I am clicking inside a different portal. ```defaultDisplayMode``` sounds good. I will also add ```blurGateDisabled``` there. K?

PeterKottas:
> yes it should be single, but you changed it to double quotes

My bad, sry got myself confused. I was making the code changes without having the node_modules installed. That's why it used some vs-code defaults instead of correct tslint settings. All good now.

packages/core/src/actions/cell/insert.ts
PeterKottas: Agreed! Cheers

PeterKottas:
ok, everything we discussed so far has been implemented.

PeterKottas:
Seeing that the 2 of us are for the most part the most active contributors/maintainers, I was curious if it would make sense to discuss our approach to closing and merging PRs. Naturally, it's the best solution for us to check each other's work as it should be, but I am sure I've previously merged in my own PRs and so have you. It simply cannot be any other way given that we are not doing this full time and therefore have periods of time where we cannot attend to this library.

I was thinking we should probably create a few simple rules on how to proceed to make our lives easiest and most productive.

1. Let's give each other 24 hours to react to a new PR.
2. If the other person doesn't react within 24 hours we should be free to close our own PRs without further "second pair of eyes" with the exception of:
a) the PR is huge (naturally this is on case to case bases, I think we are capable enough to find the right balance here)
b) the PR proposes a breaking change to the API 
c) the PR touches something that can be considered a matter of preference

If any of these are the case, we should allow further 48 hours (even if there is no response for the other person to pass). Understandably, it's preferable to communicate with each other and for instance say "I cannot look at it for the next 3 days, but please wait for me, I'll find the time during the weekend" or something along those lines. These "guidelines" are mostly meant for when one of us is unreachable for one reason or another. Also simple stuff like style/lint issues, docs or comments fixes, cross-platform issues etc. should be merged at the person's discretion without having to wait at all.

What do you think?

PeterKottas:
Could you pls take a look @macrozone ? I've got some other stuff riding on this, cheers!

packages/core/src/Provider/index.tsx
macrozone: @PeterKottas the blur should work through a portal. in fact, the bottom toolbar is a portal.

The BlurGate checks whether the click was inside the react-tree of the component.
But this has only been tested with the @react-page/editor, not with the old approach. I think i now why it is not working

macrozone:
> Seeing that the 2 of us are for the most part the most active contributors/maintainers, I was curious if it would make sense to discuss our approach to closing and merging PRs. Naturally, it's the best solution for us to check each other's work as it should be, but I am sure I've previously merged in my own PRs and so have you. It simply cannot be any other way given that we are not doing this full time and therefore have periods of time where we cannot attend to this library.
> 
> I was thinking we should probably create a few simple rules on how to proceed to make our lives easiest and most productive.
> 
> 1. Let's give each other 24 hours to react to a new PR.
> 2. If the other person doesn't react within 24 hours we should be free to close our own PRs without further "second pair of eyes" with the exception of:
>    a) the PR is huge (naturally this is on case to case bases, I think we are capable enough to find the right balance here)
>    b) the PR proposes a breaking change to the API
>    c) the PR touches something that can be considered a matter of preference
> 
> If any of these are the case, we should allow further 48 hours (even if there is no response for the other person to pass). Understandably, it's preferable to communicate with each other and for instance say "I cannot look at it for the next 3 days, but please wait for me, I'll find the time during the weekend" or something along those lines. These "guidelines" are mostly meant for when one of us is unreachable for one reason or another. Also simple stuff like style/lint issues, docs or comments fixes, cross-platform issues etc. should be merged at the person's discretion without having to wait at all.
> 
> What do you think?

I  totally agree to move fast forward. and agree with the reaction time ;-)

In this case i am fine if the `disableBlurGate` prop stays, but there are two things to consider:

- I still would like to know why it does not work properly for you, but we can investigate after we have merged it
- we should reduce the api surface, so every new prop should be challenged. Maybe we define that we "underscore" non-documented, possible future breaking props like this? `_disableBlurGate`, so that we don't have to apologize in the future ;-)

also notice that if you disable the blur gates functionality, your users can get confused with a wrong plugin-panel

macrozone:
you can merge if you want, notice my concerns above

PeterKottas:
Cool, I will proceed accordingly. 

> I still would like to know why it does not work properly for you, but we can investigate after we have merged it

Simply put, I need the plugin panel to stay open even after I click away. In my tutorial that uses https://reactour.js.org/, I am highlighting parts of the bottomToolbar and letting people click next to move to the next button explanation. The problem is, if I click next, the toolbar disappears breaking the functionality. 

> Agreed about the API surface and the props, although I must say, right now, I think we lack behind in terms of customization. The only way to achieve that is to somewhat increase the API surface. Sensibly of course.

PeterKottas:
Quick question, is all that's needed to release new version pushing a new tag? Rest is automated? I am not 100% sure. TY

Issue #805, opened 2020-05-18
Allow move in edit mode


PeterKottas:
**Is your feature request related to a problem? Please describe.**
Similarly as in https://github.com/react-page/react-page/issues/804 and for the same reasons (users finding the modes hard to follow) I am proposing to get rid of the layout mode.

**Describe the solution you'd like**
What I would like to do here is a simple fix (a couple of 1-2 liners) which would allow the dnd functionality during editing. Instead of having the overlay across the whole component, we would simply have a move icon in the top right corner. Grabbing that would allow you to move the component without having to switch modes. 

**Additional context**
This would require commenting out [this](https://github.com/react-page/react-page/blob/master/packages/core/src/components/Cell/Draggable/index.tsx#L50) and [this](https://github.com/react-page/react-page/blob/master/packages/core/src/components/Cell/Droppable/index.tsx#L37) and a few css changes for the handle and overlay.

Similarly to #804 this is not really from my head but from our experience when testing on real users. I think especially non-technical folks are having trouble grasping the concept of modes properly and it's necessarily confusing for them. That being said, understandably, this is a kind of "battle of preference" but I believe we should allow certain flexibility when it comes to this. Mode-less way of doing this is I assume no better or worse, just might be easier to grasp by certain industries.

PR #809, opened 2020-05-20
Allow resize and move in edit mode

### LINKED ISSUE #804: Allow resize in edit mode
**Is your feature request related to a problem? Please describe.**
This is something that came up during our extensive user testing. Simply put, people (non-technical mostly) find it weird and redundant to be forced to switch to resize mode in order to be able to resize columns. I tend to agree.

**Describe the solution you'd like**
I think we can allow the resize functionality to work in the edit mode.

**Describe alternatives you've considered**
Removing the resize mode altogether, however, I am not sure if that's necessary. I think we should keep it to avoid breaking changes.

**Additional context**
It is literally as easy as removing the ```this.props.isResizeMode``` from https://github.com/react-page/react-page/blob/master/packages/core/src/components/Cell/index.tsx#L100 . Optionally we could allow this functionality via a flag provided to the editor. 

@aeneasr thoughts thoughts thoughts? :) I am happy to defend this and hopeful I am not forgetting about something. I am aware this would have some performance impact but based on my preliminary tests, this appears to be negligible/not noticeable.

PeterKottas:
This resolves #804 and #805. You can check-out this functionality [here](https://screencast-o-matic.com/watch/cYhYIak5a9).


macrozone:
this is extremly cool! I love it!

PeterKottas:
Good stuff, I'll fix conflicts and merge. If there are any issues, I'll fix them as we find them. Cheers.

PR #813, opened 2020-05-20
Fixes blur gate and remove docsify dependency

PeterKottas:
There was a bug in the blur gate. Apparently you cannot return null in useEffect, also a small issue with dependency. @macrozone, could you please release this? Or preferably let me know how to release so that I don't have to bug you about this. Cheers!

### DIFF
--- a/examples/src/elements.css
+++ b/examples/src/elements.css
@@ -20,10 +20,6 @@
  *
  */
 
-.ory-cell-leaf {
-  margin: 12px;
-}
-
 .ory-plugins-content-divider {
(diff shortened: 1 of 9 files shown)


PeterKottas:
Sry for the flood of notifications @macrozone, this one is the only high priority one. Pls release when you have a chance, ty a million ;)

PeterKottas:
Ok, let's see if I can figure out how to release :)

PeterKottas:
Easier than it seemed, new version released!

Issue #803, opened 2020-05-18
when editor is inside a `display: none` the toolbar is positioned wrong


macrozone:
**Describe the bug**

When the editor is inside a a element with `display: none` it still get rendered, but the size computation from the browser will return a 0 size box. This messes with the toolbar positioning.
when you resize afterwards (and after it became visible), the position is correct again

This happens for example if you put an Editor inside a react-admin tab (in newer react-admin versions).

**Expected behavior**

position calculation is currently done on window resize and scroll. There are some newer apis to check whether an element became visible that we can use here.

PeterKottas:
I wonder if we should remove https://github.com/react-page/react-page/blob/master/packages/core/src/components/Dimensions/index.tsx altogether and replace it with something like https://github.com/ctrlplusb/react-sizeme. I am using that in one of my projects and it's working quite nicely. That being said, Not sure if it would fix this issues. Maybe this is a separate thing.

PR #820, opened 2020-06-04
fix: when editor changes from display:none to visible default ui and …

macrozone:
…resize handles are wrong

fixes https://github.com/react-page/react-page/issues/803

i fixed it by using interactionObserver. This does not work in IE11, but the editor should not be used there anyway (only in readOnly mode).

@PeterKottas I'll merge and release today, i'll guess this small change is ok

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -56,7 +56,7 @@
     "no-switch-case-fall-through": true,
     "no-trailing-whitespace": false,
     "no-unused-expression": [true, "allow-fast-null-checks"],
-    "no-use-before-declare": true,
+
     "one-line": [
       true,
       "check-catch",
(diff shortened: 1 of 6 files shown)


tslint.json
```diff
@@ -66,7 +66,7 @@
     ],
     "quotemark": [true, "single", "jsx-double"],
     "radix": true,
-    "semicolon": [true, "always"],
+    "semicolon": false,
```
PeterKottas, line 69: This is a fairly major change that would probably break a lot of code (in terms of lint). Why do we want to change it here?

PeterKottas:
Just the lint thing, rest looks good.

macrozone:
@PeterKottas I noticed that prettier and tslint are not in sync. but its true, the change is not ok here. we should add the config-prettier plugin (or even switch to eslint as this now supports typescript)

macrozone:
@PeterKottas as no PRs were open, i used the moment to switch to eslint and reformat everything according to the rules, 

i hope its now beetter

PeterKottas:
Good stuff. Then it's all good.

PR #822, opened 2020-06-05
chore: change to eslint

macrozone:
switches from tslint to eslint as this is deprecated

i also added proper prettier config to eslint, so i hope this should fix the missmatch between linting rules and prettier

maybe some rules are not perfect yet, i also needed to add some ignores in some cases, i would try to reduce these in future refactorings


PeterKottas:
Wow, you're right. Ts lint is indeed deprecated. HTF have I missed that? :D

macrozone:
@PeterKottas they annonced it long ago, but i also did not know that it finally happend until recently ;-)

PeterKottas:
Thanks for letting me know, I'll have to look into this for our repos as well!

PR #823, opened 2020-06-05
Feat: multi-language cells (experimental)

macrozone:
try it out: v0.22.0-beta.0

If you'd ask me, i would have said that i18n would belong outside of react-page. But then you'd have to accept that editors need to re-create the whole page on every language. Or developers would need to implement some copy mechanics.

On the other hand, its often much more convenient if editors could just edit one cell in different languages instead. This was surprisingly easy to implement in react-page, so i thought i would give it a go.

So here is the deal:

- you can pass optionally `languages` and a current `lang` to `<Editor lang={en} languages={[{lang: "en", label: "English"}, {lang: "de", label: "Deutsch"}]}/>`
- if you do so, you will see a language selector in the bottom toolbar, there you can select the current viewed language for the editor
- if you edit the content of a cell, you only edit the current selected language 
- if there exist no state on a cell for a given language, it will fall back to the first non-empty language
- you can also hide/show the cell per language with the Draft-switch. This would allow to further customize the differnet language versions if they are more different

Caveats:

- the state per language is stored in the tree of react-page's state, so all languages are in the same doc. if you have a lot of languages you should consider that all data has to be sent to the client. As texts compress usually nicely, this is probably neglectable when there are not many languages


macrozone:
field test looks good so far, will merge and release that soon!

fernandobandeira:
I think this is looking good, its always nice to have this builtin in the components, we can always do something about the caveat in the future so the approach should be alright for now 👍

PR #825, opened 2020-06-23
adds allowNeighbours option for layout plugins

hb90dev:
This feature is for allowing layout components to not be droppable (not be able to drop near layout)

### DIFF
--- a/packages/core/src/service/plugin/classes.ts
+++ b/packages/core/src/service/plugin/classes.ts
@@ -124,6 +124,8 @@ export type LayoutPluginExtraProps<T = any> = {
   createInitialChildren?: () => InitialChildrenDef;
 
   Component?: PluginComponentType<LayoutPluginProps<T>>;
+
+  allowNeighbours?: boolean;
 };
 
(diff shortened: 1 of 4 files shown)


packages/core/src/service/plugin/classes.ts
```diff
@@ -31,12 +31,14 @@ export type Plugins = {
   layout?: LayoutPluginConfig[];
   content?: ContentPluginConfig[];
   native?: NativeFactory;
+  missingPlugin?: PluginComponentType;
```
macrozone, line 34: can you rename it to: `MissingPluginComponent` ? to highlight that you need to define a `Component` here.

macrozone (changes requested):
one small change (sorry for the late response).

macrozone:
@hb90dev I will look at it again after the refactor.

Also there are two changes here:

- allow to specify MissingPluginComponent
- dissallow to add cells next to layout plugins

macrozone:
thhis is no obsolete and will land in 1.0.0

Issue #767, opened 2020-01-17
Sponsoring & company support


macrozone:
We at panter plan to invest more in react-page and help improving it and make it a sustainable open source project. 

We plan to:

- improve the documentation
- improve the example to show off more features
- feature it at react-meetups in zürich or similar
- move the example application as i think we have no access to the current deployment
- add better release-management (semantic-release)

On the other hand, we would like to mention us and other contributors to the readme and to the example application as companies that support that library (ORY, panter, guestbell, _your company_?). I also believe that this adds trust to the library (and gives these companies some visibility of course).

I would like to have an endorsement from other contributors here and also ask for other suggestions and other companies and individuals that want to get mentioned as either users of the libraries or active contributors. thoughts?

fernandobandeira:
Makes a lot of sense to me, we should be able to host something on github pages for the examples.

About the company, definitely, let's add that, you're the one that's dedicating more time to the lib currently so def add that.
Just keep the original authors section / formerly known.

Eventually @goodworld will be helping out more, specially on improving the UI/UX and customizations.

PeterKottas:
It sounds like a good idea. I have big plans for new features and general improvements but us (GuestBell) being in hospitality, we have more pressing issues now with the virus and all. Once the sector picks up and we can come back to development I am sure we'll be adding more stuff. I especially would like to improve the UX/UI around the drag/drop and resize functionality. Based on our user testing, it is currently somewhat unpredictable and hard to learn to use efficiently. 

In terms of mentioning companies/authors here, I am definitely happy to give my ok. I think it's the right choice to give endorsement to everybody who is contributing here. Making it more sustainable is quite a good idea, also moving the example. Right now, the link is broken which definitely doesn't look very professional/well maintained.

macrozone:
We are improving docs and examples right now, expect some improvements this week

PR #833, opened 2020-08-24
Support

PeterKottas:
This is a followup to https://github.com/react-page/react-page/issues/767. It appears we all agreed to go forward with it but didn't implement it. I've added our credentials with 2 sections.

"In the wild" Should link directly to places where the react page is used. I could imagine this to be end products as well as popular codepens and so on.

"Sponsored" is for companies and individuals who invest time money or other efforts. All should be sorted alphabetically. 

Feel free to check and add your credentials as well gents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Original Library created by [@aeneasr](https://github.com/aeneasr) [@ory](https:
 
 ## In the wild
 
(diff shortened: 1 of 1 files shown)


macrozone:
cool! thx!

I will have a hack week soon (later september). i plan to give it some more love then!

PeterKottas:
@macrozone Ty man, sure thing.

PR #860, opened 2020-10-06
WIP: Big refactor

macrozone:
this will be huge! This will drastically simplify development , should improve performance and should even reduce the bundle size. At some point we even be able to remove redux

- [x] rework new hooks and add new ones
- [x] remove native plugin (will be readded in a future directly on plugins)
- [x] refactor Content to use hooks
- [x] refactor Layout to use hooks
- --> unify Content & Layout
- [x] refactor Rows to use hooks
- [x] refactor plugins (remove useless plugin wrapper class)
- [ ] rename create-plugin-materialui  and unify with existing plugins
- [ ] document new hooks (for developers)


macrozone:
ill continue in smaller PRs now on beta

PR #911, opened 2021-02-23
Docs update 1

zehawki:
The first of many documentation updates :-)


docs/editor.md
```diff
@@ -33,37 +34,38 @@ const App = () => {
 
 ### `value: Value`
 
-this is the value that the editor should display. You will usually load that from a database.
-It has the type `Value`, but should generally considered to be "opaque", so don't rely on its internal structure.
+This is the content that the editor will display. The data for this prop may come from any source, eg a file or database or API etc. It is a JSON representation of the entire content that should be rendered, including the external visible information, and internal information required to render the information - eg ids, versions, plugin info, etc.
```
macrozone, line 37: really nice! Maybe omit or reword the "including the external visible information and internal information" part, personally I don't really get that and i think its more confusing than helpful.

docs/editor.md
```diff
@@ -33,37 +34,38 @@ const App = () => {
...
+
+This information is "opaque", and is not meant to be edited directly under normal circumstances, so don't rely on its internal structure.
+
+It does not contain any presentation aspects of the data, ie no CSS is stored.
 
 ### `onChange: (newValue: Value) => void`
 
-this function is called whenever the editor has new data. You would usually store the new data back to the database in this function.
-This is not required when `readOnly` is `true`
+A callback function whenever the editor has new data. This is not required when `readOnly` is `true`
```
macrozone, line 45: Maybe readd a sentence about storing the data

docs/editor.md
```diff
@@ -33,37 +34,38 @@ const App = () => {
...
 
 ### `readOnly: boolean`
 
-if set to `true`, the content cannot be edited. You need to set this if you want to just isplay your content.
-Any code that is only used for editing won't be loaded if you using webpack or similar bundlers, so you don't have to worry about bundle size!
+If set to `true`, the content cannot be edited. Set this when using the editor to display the content. The code that is used for editing isn't loaded in this case, and hence there is a reduction in the budle size if using webpack or similar bundlers.
```
macrozone, line 49: much better!

docs/editor.md
```diff
@@ -33,37 +34,38 @@ const App = () => {
...
 ### `cellPlugins: CellPlugin[]`
 
-an array of `CellPlugin`s that can be used in this editor.
+An array of `CellPlugin`s that can be used in this editor.
 
-We provide both some ready-to-use plugins and an api to create custom cell plugins to display anything you want.
+'react-page' comes with some inbuilt plugins and a superb extensible system to create new plugins for displaying anything.
 
-You will usually need the [`slate`](/slate.md) plugin for rich text editing and some custom plugins.
+At the very least, the [`slate`](/slate.md) inbuilt plugin for rich text editing is needed.
```
macrozone, line 63: I would also argue that you will always want that plugin, but there are users out there of the library who actually do not use it ;-)

docs/editor.md
```diff
@@ -104,3 +106,63 @@ it will only show the (+) button to add new cells when it has less than `maxChil
...
+          "rows": [
+            
+          ],
+          "inline": null
+        }
+      ]
+    }
+  ]
+}
+```
```
macrozone, line 168: i think this is a nice addition. Typescript users might also want to look into the `Value` type.

docs/slate.md
```diff
@@ -12,7 +12,7 @@ $ npm i --save @react-page/plugins-slate
 
 <p>
   <figure align="center">
-    <img alt="Text editing plugin" src="./images/text-editing-plugin.gif"><br>
+    <img alt="Text editing plugin" src="../docs-images/text-editing-plugin.gif"><br>
```
macrozone, line 15: thx, I did some linking magic to make the docs run inside the example nextjs app, so i broke some links.

I'll have to check whether it works after deployment.

macrozone:
thank you very much for the improvements! I added some comments, but its good enough to merge (and much better than my wording anyway).

Maybe prefix your commits with `docs:` so that it will be properly handled by semantic release and added to the changelog. Currently, the demo project will only be updated when a new version will be published, which is only the case for `fix:` or `feat:` commits. I plan to also do new publishs on `doc:` commits soon

macrozone (approved):

docs/editor.md
zehawki: I'll try rewording it better in the next version. "external visible information" = what the viewer sees. "internal information" =the meta data that controls the rendering

docs/editor.md
zehawki: I left that out on purpose. onChange is not equal to storing the data. When someone uses the editor, onChange will fire literally 100s of times or more. The onChange should update an internal variable, which is the standard controlled mode of react. At the end the user may discard changes, or save - and at that point the data would be stored somewhere, file, API etc, but that has nothing to do with the onChange callback.

zehawki:
> Maybe prefix your commits with docs: so that it will be properly handled by semantic release and added to the changelog. 

Got it, will do

zehawki:
@macrozone I'm a bit confused as to whats going on, on this PR and the best way to proceed. Did you get a notification for the next commit I did on the docs earlier today? Should I continue pushing into this branch and leave this PR open , or close and file a new PR each time

macrozone:
> @macrozone I'm a bit confused as to whats going on, on this PR and the best way to proceed. Did you get a notification for the next commit I did on the docs earlier today? Should I continue pushing into this branch and leave this PR open , or close and file a new PR each time

its ok to push several commits concerning the same topic. Please be patient, i review it if i find time

docs/editor.md
macrozone: this is absolutly correct, yes.

zehawki:
@macrozone something has gone a bit wrong in the merge, see https://react-page.github.io/docs/#/editor?id=_2-simple-example-with-image at the bottom. The part added by @avoskresensky got merged into the code block...

image

macrozone:
thx, will fix it

PR #914, opened 2021-02-25
feat(ui): ability to override BottomToolbar component

### LINKED ISSUE #909: Make BottomToolbar optional
We use our own set of plugins with our own plugin controls. In version 0.x we had an ability to display our own toolbox, but we seem to have lost that ability in 1.x.

BottomToolbar is now always mounted inside the PluginComponent -- even when plugin.controls is null.

We suggest to mount BottomToolbar conditionally -- only when plugin.controls is set. That would allow us to have our own version of it mounted in the Renderer component.

avoskresensky:
## Ability to override BottomToolbar component

Add an ability to override certain internal components starting with BottomToolbar.

## Types of changes

## Checklist

## Closing issues

closes #909

### DIFF
--- a/examples/components/BottomToolbar/CollapseButton.tsx
+++ b/examples/components/BottomToolbar/CollapseButton.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import { IconButton, Tooltip } from '@material-ui/core';
+import IconCollapse from '@material-ui/icons/KeyboardArrowDown';
+import IconRestore from '@material-ui/icons/KeyboardArrowUp';
+
+interface CollapseButtonProps {
+  collapsed: boolean;
+  toggleCollapsed(): void;
+}
+
+const CollapseButton: React.FC<CollapseButtonProps> = ({
+  collapsed,
+  toggleCollapsed,
+}) => {
+  return (
+    <Tooltip title={collapsed ? 'Restore Panel' : 'Collapse Panel'}>
+      <IconButton
+        onClick={() => toggleCollapsed()}
+        aria-label="delete"
+        color="default"
+      >
+        {collapsed ? <IconRestore /> : <IconCollapse />}
+      </IconButton>
+    </Tooltip>
+  );
+};
+
+export default React.memo(CollapseButton);
--- a/examples/components/BottomToolbar/Content.tsx
+++ b/examples/components/BottomToolbar/Content.tsx
@@ -1,0 +1,68 @@
+import {
+  Avatar,
+  Grid,
+  IconButton,
+  Tooltip,
+  Typography,
+} from '@material-ui/core';
+import FormatSize from '@material-ui/icons/FormatSize';
+import React from 'react';
+import { usePluginOfCell } from '@react-page/editor';
+import Tools from './Tools';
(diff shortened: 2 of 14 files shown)


macrozone:
looks exactly like i thought it would!

One small thing: we have to be very careful about rerenders (see comment)

Another thing: can you add a new page in examples/pages/examples where we have  a custom toolbar?

avoskresensky:
@macrozone, could you point me to that comment you're referring to? I'm not seeing anything in the "files changed" section.

What do you have in mind re the example? Our own use case is `components={BottomToolbar: ()=>null}`.
Or do you want something with a useful alternative toolbar?

macrozone (changes requested):

macrozone:
> @macrozone, could you point me to that comment you're referring to? I'm not seeing anything in the "files changed" section.

seems i had to submit the review first

> What do you have in mind re the example? Our own use case is `components={BottomToolbar: ()=>null}`.
> Or do you want something with a useful alternative toolbar?

would be nice, yes. Maybe show your example. I think setting it to null is probably not a good solution at all, you then cannot display any controls. Notice that you should avoid mounting custom controls in the Renderer itself as this might lead to problems with readOnly / static output as increased bundle size.

instead you should replace BottomToolbar with something that renders its children (e.g. in a modal, drawer, or similar).

See the code that renders the bottom toolbar:

Bildschirmfoto 2021-02-26 um 11 53 21

you see that it will render whatever controls the plugin define

avoskresensky:
@macrozone, I've made the requested mods and added an example.

The example demonstrates adding an additional control (collapse button) to the drawer. The majority of other controls were reimplemented via copy-pasta with public dependencies. Added an extra public export to the main package so I don't have to reimplement ui/ThemeProvider.

macrozone:
@avoskresensky nice! its ok if i finish it a bit? I think its a good idea to replicate the BottomToolbar, but i want to highlight that this is only as an example, you usually do not have to do that.

macrozone:
@avoskresensky i further refactored the BottomToolbar and now export a lot of internal components for easier customization.

This allows us to customize the ui more easily without recreating every internal component.

At some point in the future we can abstract away material-ui more easily so that we can replace Buttons, Drawer, etc. with custom elements without losing the already existing functionality

macrozone:
its already merged, but i could not update this pr, so checkout the changes on master!

avoskresensky:
@macrozone , awesome stuff. Much nicer this way.

One comment though: 

https://github.com/react-page/react-page/blob/master/packages/editor/src/core/components/Cell/PluginComponent/index.tsx#L87

It would be nicer if that element got a named class and the styles got moved into the CSS. Would be easier to customize if the custom BottomToolbar requires something else.

macrozone:
> @macrozone , awesome stuff. Much nicer this way.
> 
> One comment though:
> 
> https://github.com/react-page/react-page/blob/master/packages/editor/src/core/components/Cell/PluginComponent/index.tsx#L87
> 
> It would be nicer if that element got a named class and the styles got moved into the CSS. Would be easier to customize if the custom BottomToolbar requires something else.

good idea, will do that.

Issue #918, opened 2021-03-02
Cell spacing (grid gutters)


avoskresensky:
Add an ability to have cell spacing (grid gutters).

Currently the only way to add something like that would be via .react-page-cell-inner-leaf padding (as it is done in the examples). This solution is very simple and straightforward, but has the following problems:
- padding is added to the outside perimeter of the grid (including nested grids)
- if a plugin has any chrome (content on a card, a border etc) this starts to look weird
- won't work on backgrounds sitting in a row next to each other or any other layout plugins

We would like to be able to specify cellSpacing as an Editor prop (`number | number[]`). Ideally we would also like to be able to override this property inside a layout plugin to have different cell spacing in the nested grid.

The flexboxgrid that the CSS is based on had a part of that functionality built in via --gutter-width property. Applying padding to .react-page-cell and a negative margin to .react-page-row seems to do the trick for horizontal gutters.

Vertical gutters are trickier and require some additional checks, but the general idea is still the same: padding + negative margin.

We would like to propose a fairly straightforward implementation:
https://github.com/react-page/react-page/compare/master...avoskresensky:cell-spacing-v1

This is just a demonstration of the approach and it lacks certain features like HTMLRenderer modifications, separate horizontal/vertical settings for cellPadding, some drag-and-drop guides are misaligned etc.

If the general idea is acceptable, we can finalize the PR.

macrozone:
@avoskresensky its already possible to set the cell style for a CellPlugin:

```
 cellStyle: {
    padding: 0,
  },
```

so you can "globally" set the padding with css and override it for certain cellPlugins

https://github.com/react-page/react-page/blob/master/packages/plugins/layout/background/src/createPlugin.tsx#L31
https://github.com/react-page/react-page/blob/master/examples/pages/examples/constraints.tsx#L30

avoskresensky:
@macrozone, cell padding is not gutters. In order to become gutters it needs to be compensated around the grid. That's what the negative margin is for.

Also, as I mentioned above there is a cosmetic matter of being able to highlight just the contents of the cell -- not the cell plus its adjoining "gutters".

macrozone:
you can also set negative margin with `cellStyle` its applied to the outer div.

avoskresensky:
@macrozone, negative margin needs to be applied on enclosing element -- a row in case of horizontal gutters. See the flexgrid CSS. With vertical gutters the compensation is even trickier.

avoskresensky:
[...]

PR #919, opened 2021-03-02
fet(ui): cell spacing

avoskresensky:
## Proposed changes

Add `cellSpacing` prop to the Editor where they can specify the sizes of grid gutters (vertical and horizontal).

## Types of changes

## Checklist

## Closing issues

#918

## Further comments

See examples/pages/examples/cellSpacing.tsx for the demonstration of the new functionality.

The main layout change is the split of .react-page-cell into two divs. The outer div gets all the flexgrid classes, inline classes (floats) and it's the div that gets all the paddings that constitute the gutters. The inner div retains all the rest of the classes and is the div that gets highlighted when a cell is selected.

Horizontal padding compensation is done by applying a negative margin to the .react-page-row class.

Vertical compensation is only needed for
- nested grids (layout plugins) and is applies to the new div that encloses the plugin's children (excluding the InsertNew component).
- the entire editor -- applied inside Editable/Inner component

### DIFF
--- a/examples/pages/examples/cellSpacing.tsx
+++ b/examples/pages/examples/cellSpacing.tsx
@@ -19,8 +19,8 @@ interface CellSpacingState {
   readOnly?: boolean;
   outlineEditor?: boolean;
   outlineCells?: boolean;
-  cellSpacingX?: string;
-  cellSpacingY?: string;
+  cellSpacingX?: number;
+  cellSpacingY?: number;
 }
 
 // Remove samples stylesheet cell padding and optionally outline cell and editor boundaries.
@@ -44,8 +44,8 @@ export default function CellSpacing() {
     value: cellSpacingDemo,
     outlineEditor: true,
     outlineCells: true,
-    cellSpacingX: '10',
-    cellSpacingY: '10',
+    cellSpacingX: 10,
+    cellSpacingY: 10,
   });
 
   const style = React.useMemo(
@@ -59,8 +59,8 @@ export default function CellSpacing() {
       outlineEditor: { type: 'boolean', title: 'Outline the Editor in Green' },
       outlineCells: { type: 'boolean', title: 'Outline Cells in Red' },
       readOnly: { type: 'boolean', title: 'Read Only' },
-      cellSpacingX: { type: 'string', title: 'Horizontal Cell Spacing' },
-      cellSpacingY: { type: 'string', title: 'Vertical Cell Spacing' },
+      cellSpacingX: { type: 'number', title: 'Horizontal Cell Spacing' },
+      cellSpacingY: { type: 'number', title: 'Vertical Cell Spacing' },
     },
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   } as any);
@@ -107,9 +107,7 @@ export default function CellSpacing() {
           cellPlugins={cellPlugins}
           value={state.value}
           onChange={(value) => setState((s) => ({ ...s, value }))}
-          // let's also test our ability to survive non-numbers
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          cellSpacing={[state.cellSpacingX as any, state.cellSpacingY as any]}
+          cellSpacing={{ x: state.cellSpacingX, y: state.cellSpacingY }}
         />
       </div>
     </PageLayout>
--- a/examples/pages/examples/cellSpacing.tsx
+++ b/examples/pages/examples/cellSpacing.tsx
@@ -1,53 +1,100 @@
-import React, { useState } from 'react';
-
-// The editor core
-import Editor, { Value } from '@react-page/editor';
-
-// import the main css, uncomment this: (this is commented in the example because of https://github.com/vercel/next.js/issues/19717)
-// import '@react-page/editor/lib/index.css';
-
-// The rich text area plugin
+import React from 'react';
+import Editor, {
+  Value,
+  AutoForm,
+  AutoFields,
+  makeUniformsSchema,
+} from '@react-page/editor';
 import slate from '@react-page/plugins-slate';
-// image
 import image from '@react-page/plugins-image';
-// background
 import background from '@react-page/plugins-background';
 import PageLayout from '../../components/PageLayout';
-
-// Stylesheets for the rich text area plugin
-// uncomment this
-//import '@react-page/plugins-slate/lib/index.css';
-
-// Stylesheets for the imagea plugin
-//import '@react-page/plugins-image/lib/index.css';
 
 // Define which plugins we want to use.
 const cellPlugins = [slate(), image, background({})];
 
-// Remove samples stylesheet cell padding, highlight cell and editor boundaries to demonstrate the cell spacing.
-const styles = `
-.react-page-cell-inner-leaf {
-  padding: 0;
+interface CellSpacingState {
+  value?: Value;
+  outlineEditor?: boolean;
+  outlineCells?: boolean;
+  cellSpacingX?: string;
+  cellSpacingY?: string;
 }
-.react-page-cell-inner > div {
-  outline: 1px solid red;
-}
-.react-page-editable {
-  outline: 1px solid green;
-}
-`;
 
-export default function SimpleExample() {
-  const [value, setValue] = useState<Value>(null);
+// Remove samples stylesheet cell padding and optionally outline cell and editor boundaries.
+function styles(state: CellSpacingState) {
+  let styles = '.react-page-cell-inner-leaf { padding: 0; }';
+
+  if (state.outlineCells) {
+    styles += '.react-page-cell-inner > div { outline: 1px solid red; }';
+  }
+
+  if (state.outlineEditor) {
+    styles += '.react-page-editable { outline: 1px solid green; }';
+  }
+
+  return styles;
+}
+
+export default function CellSpacing() {
+  const [state, setState] = React.useState<CellSpacingState>({
+    cellSpacingX: '10',
+    cellSpacingY: '10',
+  });
+
+  const style = React.useMemo(
+    () => <style dangerouslySetInnerHTML={{ __html: styles(state) }} />,
+    [state.outlineCells, state.outlineEditor]
+  );
+
+  const schema = makeUniformsSchema({
+    type: 'object',
+    properties: {
+      outlineEditor: { type: 'boolean', title: 'Outline the Editor in Green' },
+      outlineCells: { type: 'boolean', title: 'Outline Cells in Red' },
+      cellSpacingX: { type: 'string', title: 'Horizontal Cell Spacing' },
+      cellSpacingY: { type: 'string', title: 'Vertical Cell Spacing' },
+    },
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  } as any);
 
   return (
     <PageLayout>
-      <style dangerouslySetInnerHTML={{ __html: styles }} />
+      {style}
+
+      <div
+        style={{
+          padding: '20px',
+          marginBottom: '20px',
+          outline: '1px solid #E0E0E0',
+        }}
+      >
+        <AutoForm
+          model={state}
+          autosave={true}
+          schema={schema}
+          onSubmit={(val: Partial<CellSpacingState>) =>
+            setState((s) => ({ ...s, ...val }))
+          }
+        >
+          <div
+            style={{
+              columnCount: 3,
+              columnGap: 48,
+            }}
+          >
+            <AutoFields />
+          </div>
+        </AutoForm>
+      </div>
+
       <Editor
         cellPlugins={cellPlugins}
-        value={value}
-        onChange={setValue}
-        cellSpacing={20}
+        value={state.value}
+        onChange={(value) => setState((s) => ({ ...s, value }))}
+        // let's also test our ability to survive non-numbers
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        cellSpacing={[state.cellSpacingX as any, state.cellSpacingY as any]}
       />
     </PageLayout>
   );
--- a/examples/pages/examples/cellSpacing.tsx
+++ b/examples/pages/examples/cellSpacing.tsx
@@ -1,0 +1,54 @@
+import React, { useState } from 'react';
+
+// The editor core
+import Editor, { Value } from '@react-page/editor';
+
+// import the main css, uncomment this: (this is commented in the example because of https://github.com/vercel/next.js/issues/19717)
+// import '@react-page/editor/lib/index.css';
+
+// The rich text area plugin
+import slate from '@react-page/plugins-slate';
+// image
+import image from '@react-page/plugins-image';
+// background
+import background from '@react-page/plugins-background';
+import PageLayout from '../../components/PageLayout';
+
+// Stylesheets for the rich text area plugin
+// uncomment this
+//import '@react-page/plugins-slate/lib/index.css';
+
+// Stylesheets for the imagea plugin
+//import '@react-page/plugins-image/lib/index.css';
+
+// Define which plugins we want to use.
+const cellPlugins = [slate(), image, background({})];
+
+// Remove samples stylesheet cell padding, highlight cell and editor boundaries to demonstrate the cell spacing.
+const styles = `
+.react-page-cell-inner-leaf {
+  padding: 0;
+}
+.react-page-cell-inner > div {
+  outline: 1px solid red;
+}
+.react-page-editable {
+  outline: 1px solid green;
+}
+`;
+
+export default function SimpleExample() {
+  const [value, setValue] = useState<Value>(null);
+
+  return (
+    <PageLayout>
+      <style dangerouslySetInnerHTML={{ __html: styles }} />
+      <Editor
+        cellPlugins={cellPlugins}
+        value={value}
+        onChange={setValue}
+        cellSpacing={20}
+      />
+    </PageLayout>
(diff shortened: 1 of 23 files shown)


packages/editor/src/core/Provider/index.tsx
```diff
@@ -161,6 +163,7 @@ const Provider: React.FC<ProviderProps> = ({
     JSON.stringify(childConstraints ?? {}), // its an object, we prevent unnecessary rerenders by stringify it
     ...Object.keys(components ?? {}),
     ...Object.values(components ?? {}), // minimize unnecessary rerenders by forcing shallow comparison of "components" object
+    cellSpacing,
```
macrozone, line 166: if its a number, this works, but it might lead to rerenders if it is an array or object (if we e.g. want to allow to have dfifferent vertical and horizontal padding)

JSON.stringify might be better here

packages/editor/src/core/components/Cell/PluginComponent/index.tsx
```diff
@@ -73,7 +76,16 @@ const PluginComponent: React.FC<{ nodeId: string; hasChildren: boolean }> = ({
           }}
         >
           {Renderer ? (
-            <Renderer {...componentProps}>{children}</Renderer>
+            <Renderer {...componentProps}>
+              {hasChildren ? (
+                <div style={{ marginBottom: `${-cellSpacing}px` }}>
+                  {children}
+                </div>
```
macrozone, line 83: it might be better to just do that in the parent component. This way, you don't have to pass the insertAllowed prop.

so this new wrapper would be defined in the parent component and this component here can be left untouched

packages/editor/src/core/components/Cell/index.tsx
```diff
@@ -107,39 +109,51 @@ const Cell: React.FC<Props> = ({ nodeId, measureRef }) => {
 
   return (
     <div
-      ref={ref}
+      style={{
+        padding: `0 ${cellSpacing / 2}px`,
+        paddingBottom:
+          (!hasChildren || hasPlugin) && !inline ? `${cellSpacing}px` : null,
```
macrozone, line 115: hasChildren is a bit tricky. We unified layout and content plugins where a "layout" plugin is just a cell plugin that renders children.

Problem is that `hasChildren` does not tell you the difference between a plugin that can't have children and one that can have children but does not have any at the moment (e.g. an empty background). That was a slight oversight of mine.

this might lead to a padding jump here when you add the first children.

packages/editor/src/core/components/Cell/index.tsx
```diff
@@ -107,39 +109,51 @@ const Cell: React.FC<Props> = ({ nodeId, measureRef }) => {
...
         gridClass({
           isEditMode,
           isPreviewMode,
           size,
         }),
         {
           'react-page-cell-has-inline-neighbour': hasInlineNeighbour,
+          [`react-page-cell-inline-${inline || ''}`]: inline,
+        }
+      )}
```
macrozone, line 127: i seldom use inline cells, can't really tell if this change has any side effect.

packages/editor/src/core/components/Cell/index.tsx
```diff
@@ -107,39 +109,51 @@ const Cell: React.FC<Props> = ({ nodeId, measureRef }) => {
...
+    >
+      <div
+        ref={ref}
+        className={classNames('react-page-cell', {
```
macrozone, line 131: i have to check if this additional div has any side effect

packages/editor/src/core/components/Editable/Inner/index.tsx
```diff
@@ -84,10 +85,16 @@ const Inner: React.FC = () => {
       style={{ minHeight: 480 }}
       className={'react-page-editable react-page-editable-mode-' + mode}
     >
-      {rowIds.length > 0
-        ? rowIds.map((id) => <Row nodeId={id} key={id} />)
-        : null}
-      {insertAllowed ? <InsertNew /> : null}
+      <div
+        style={{
+          marginBottom: mode !== 'preview' ? null : `${-cellSpacing}px`,
```
macrozone, line 90: why is the spacing removed on preview mode?

packages/editor/src/core/components/Row/index.tsx
```diff
@@ -57,7 +64,11 @@ const Row: React.FC<{ nodeId: string }> = ({ nodeId }) => {
             rowHasInlineChildrenPosition
           ),
         })}
-        style={{ position: 'relative', borderColor: 'red' }}
+        style={{
+          position: 'relative',
+          borderColor: 'red',
```
macrozone, line 69: the red border can be removed, i guess that as a misstake

macrozone:
@avoskresensky thank you for your effort! if you can resolve some discussions i can merge it into beta first, then its easier to try it out in different apps to see if there is an unwanted side effect.

maybe you can also think about a more general solution if you want

packages/editor/src/core/Provider/index.tsx
avoskresensky: I was going to do something like `Array.isArray(cellSpacing) ? ...cellSpacing : cellSpacing` but it looks like TS doesn't recognize `Array.isArray` as a type guard. JSON.stringify is going to be shorter.

packages/editor/src/core/components/Cell/PluginComponent/index.tsx
avoskresensky: Will do.

packages/editor/src/core/components/Cell/index.tsx
avoskresensky: hasChildren check here is done for the sake of this situation:

image

The jumbo cell on the right has children (three rows), but it should not have its own vertical padding as it has been taken care of by the individual cells inside.

hasPlugin check simply makes sure we add padding to layout plugins, whether they have children or not.

image

The first background is empty (I left the internal padding on so it doesn't collapse) and will receive its gutter at the bottom. The second background has children (one cell), but being a standalone cell it should still receive its gutter which it does.

This logic doesn't really depend on whether we're nested (inside a layout plugin) or at first layout level.
The compensation is happening elsewhere.

packages/editor/src/core/components/Cell/index.tsx
avoskresensky: These classes declare various `float` styles and must be assigned to the top level div or the floats won't work. That's why these two classes were separated from the rest and moved to the outer div.

packages/editor/src/core/components/Editable/Inner/index.tsx
avoskresensky: Oops. I though I fixed that. The idea is simply to exclude the InsertNew component from the compensation box.
The right way of doing that is the same as in the Inner cell: wrap a compensating div around the children (if any) and exclude InsertNew.

avoskresensky:
@macrozone , pushed some changes that ...

- address the discussions
- add support for separate horiz and vert spacing
- fix the outstanding issues with the placement of dnd placeholders (some shallow layout changes there)
- beef up the sample page

Things left to do:

- HTMLRenderer
- cellSpacing override in layout plugins

macrozone:
@avoskresensky can you ping me when you're done (the two remaining things)? i think it looks already solid now and i see that you know what you are doing :-)

avoskresensky:
@macrozone, I think that's all she wrote.

Three items to discuss though:

1. Plugins are now enclosed in a <div> with display=flex. That gets rid of various abnormalities caused by margin collapsing logic of display=block. As a result, there is one notable discrepancy with the previous behavior with cell padding off. It happens with inline cells only. This is how it used to work before:

image

The bottom margin of a Slate paragraph is bleeding through the cell boundary and cell highlighting doesn't include the entire box.
Now with no margin collapsing it looks like this:

image

I think it's only for the better.
If you get rid of the inline picture plugin, "before" and "now" will look the same.

2. If you try comparing the sample page layout in readOnly and editable modes with red outlines on, you'll notice that some plugin boxes differ in size. That's not a regression -- that happens because some divs have height=100% in the editable mode and none of that in HTMLRenderer. Not sure if that's a problem.

3. A minor one, but it bugs me. I'm not sure if the sequence I chose for the cellSpacing prop is the right one. It's [x, y] at the moment which makes sense from the regular math standpoint, but is the opposite of how it goes in CSS.

examples/pages/examples/cellSpacing.tsx
```diff
@@ -0,0 +1,117 @@
...
+          value={state.value}
+          onChange={(value) => setState((s) => ({ ...s, value }))}
+          // let's also test our ability to survive non-numbers
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          cellSpacing={[state.cellSpacingX as any, state.cellSpacingY as any]}
+        />
+      </div>
+    </PageLayout>
+  );
+}
```
macrozone, line 117: thats a nice example

examples/pages/examples/cellSpacing.tsx
```diff
@@ -0,0 +1,117 @@
...
+        style={{ display: 'flex', flexDirection: 'column' }}
+      >
+        <Editor
+          readOnly={state.readOnly}
+          cellPlugins={cellPlugins}
+          value={state.value}
+          onChange={(value) => setState((s) => ({ ...s, value }))}
+          // let's also test our ability to survive non-numbers
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          cellSpacing={[state.cellSpacingX as any, state.cellSpacingY as any]}
```
macrozone, line 112: don't like the any here

packages/editor/src/core/components/hooks/options.tsx
```diff
@@ -60,3 +60,27 @@ export const useConfiguredCellPlugin = (pluginId: string) => {
...
+) => {
+  const options = useOptions();
+  return (({ children }) => (
+    <OptionsContext.Provider
+      value={{ ...options, cellSpacing: [cellSpacingX, cellSpacingY] }}
+    >
+      {children}
+    </OptionsContext.Provider>
+  )) as React.FC<unknown>;
+};
```
macrozone, line 86: that looks a bit complicted.

Should cellspacing override the spacing options for the whole subtree?

if not, you could change `useCellSpacing` to take a nodeId and it would read the plugin definition of that nodeId to check whether it defines custom cellSpacing

I am also not a fan of returning a Component in a hook. You could directly use `OptionContext.Provider`, thats simpler. The above hook has the additional problem of returning a new Component on every render, which will break memoization of components and lead to rerenders

packages/editor/src/renderer/HTMLRenderer.tsx
```diff
@@ -57,6 +95,10 @@ const HTMLCell: React.FC<
     const cellStyle = getCellStyle(plugin);
     const Provider = plugin.Provider ?? DefaultProvider;
     const data = getCellData(cell, lang);
+    const pluginCellSpacing = getPluginCellSpacing(plugin, data);
+    const [pluginCellSpacingX, pluginCellSpacingY] = pluginCellSpacing
+      ? normalizeCellSpacing(pluginCellSpacing)
```
macrozone, line 100: you could `getPluginCellSpacing` alsways normalizing it.

macrozone:
i really love this and the precision you want to have!

i am a bit deterred by the complexity but most things make sense to me. Maybe rethink the OptionContext.Provider hook a bit (added comment above).

I also pushed a small conflicting change to master, but that should be easy to rebase

examples/pages/examples/cellSpacing.tsx
```diff
@@ -0,0 +1,117 @@
...
+  );
+
+  const schema = makeUniformsSchema({
+    type: 'object',
+    properties: {
+      outlineEditor: { type: 'boolean', title: 'Outline the Editor in Green' },
+      outlineCells: { type: 'boolean', title: 'Outline Cells in Red' },
+      readOnly: { type: 'boolean', title: 'Read Only' },
+      cellSpacingX: { type: 'string', title: 'Horizontal Cell Spacing' },
+      cellSpacingY: { type: 'string', title: 'Vertical Cell Spacing' },
```
macrozone, line 63: use type `integer` or `number` instead of strings

avoskresensky: @macrozone , I want them to be able to give us random strings. I'm checking how we can survive the stuff that will convert to NaNs.

packages/editor/src/renderer/HTMLRenderer.tsx
avoskresensky: @macrozone, getPluginCellSpacing can and should return undefined. I'm checking for it a few lines down. This sequence looks a bit odd but it's due to the fact that hooks can't (or rather shouldn't) be executed conditionally.

examples/pages/examples/cellSpacing.tsx
avoskresensky: @macrozone, I we're to test our ability to survive random strings in cellSpacing prop, we need some way of shutting up TS errors. This one seems to be the shortest.

packages/editor/src/core/components/hooks/options.tsx
avoskresensky: @macrozone
Yes, the idea is that a layout plugin should be able to override cell spacing in the entire subtree of its children. It's either a provider or passing settings down in props like it's done in HTMLRenderer.

The idea behind not using OptionContext.Provider in Cell/Inner explicitly was to hide that implementation detail inside the hooks file where that Context is declared. For the cell that uses it it's just "a provider" that does the trick -- it doesn't have to concern itself which context exactly is responsible and how to construct it.

I agree with the rerender observation. The component should memoized.
Let me know if you want me to create it explicitly inside Cell/Inner or just fix the hook.

avoskresensky: @macrozone , to be completely clear, my idea of a better component returning hook is something like that:

```
const [Provider, value] = useCellSpacingProvider(x, y);
...
return (
  ...
   <Provider value={value}>
      {whatever}
   </Provider>
  ...
)
```

`Provider` is an opaque provider (the hook will be returning `OptionContext.Provider`) while `value` is a memoized opaque value. Cell/Inner is just using those without concerning itself with the particulars of the implementation.

examples/pages/examples/cellSpacing.tsx
macrozone: you mean em and rem and sort? that makes sense.

You can also check whether React.CssProperties["paddingLeft"] or so has a more precise definition for paddings

packages/editor/src/core/components/hooks/options.tsx
macrozone: i think memoizing the component is fine and if it works flaweless, we can do it like that.

i just have the feeling that this is not that idiomatic, but thats probably because i have rarely seen or thought about a hook returning a Component (hooks returning elements on the other hand are quite common). 

i agree with your reasoning though that its an implementation detail that can be hidden in the hook. So its resolved for me

examples/pages/examples/cellSpacing.tsx
macrozone: ah no, you do math with it (e.g. -cellSpacingX), so it has to be a number. see coment above

examples/pages/examples/cellSpacing.tsx
macrozone: make it stable for "missuse " (passing wrong type) is fine.  But maybe thats more a topic for testing and less for the example as it might confuse.

macrozone:
> 2\. A minor one, but it bugs me. I'm not sure if the sequence I chose for the cellSpacing prop is the right one. It's [x, y] at the moment which makes sense from the regular math standpoint, but is the opposite of how it goes in CSS.

i would align it with css, thats more what people expect. Else an object would solve the confusion as well (or tupel labels for ts users)

macrozone:
> I think it's only for the better.
> If you get rid of the inline picture plugin, "before" and "now" will look the same.

agree

> 1. If you try comparing the sample page layout in readOnly and editable modes with red outlines on, you'll notice that some plugin boxes differ in size. That's not a regression -- that happens because some divs have height=100% in the editable mode and none of that in HTMLRenderer. Not sure if that's a problem.

The 100% is to make it easier to drag and drop cells in edit mode. The inner plugin component div should have the same size though (unless a user has set 100% height on the plugin component's div as well, but i think that makes not sense in most cases.

i would not worry about that at the moment. If we see a or major unexpected difference, we can revisit

avoskresensky:
@macrozone , all discussions have been addressed. Please take a look.

macrozone:
@avoskresensky thank you for your contribution!

i merged it now into beta, that will create a new version under the "beta" tag on npm.

the reason for this is to make it easier to test it for regressions in different projects before we release it into "latest".

zehawki:
The file examples/pages/examples/cellSpacing.tsx is missing a `<SubmitField />` isnt it, under `<AutoForm`

zehawki:
In any case, I can't get the cellSpacing prop to take effect. Has this been released? I upgraded to v1.7.2

macrozone:
@zehawki this is on 2.0.0-beta release, as it might be slightly breaking

zehawki:
Ah, ok, Since [1.5.0](https://github.com/react-page/react-page/releases/tag/v1.5.0) was released I though that got merged into the latest. Really looking forward to this update.

macrozone:
@zehawki yeah, sorry about that, this is normally the case, but i noticed later that it might break the layout in certain cases, so i bumbed the major version number.

btw. i release it soon on "latest" (so no longer "beta")

PR #926, opened 2021-03-09
fix(inline): fix inline layout

avoskresensky:
## Proposed changes

Revert to using display:block type of layout in case of inline cells.

## Types of changes

## Checklist

## Closing issues

#918


avoskresensky:
@macrozone, this should take care of inline cell layout problems.

macrozone:
thx!

Issue #865, opened 2020-10-14
Providing translations for Editor and its plugins


DcBD:
# @react-page `translations`

## The main concept of defining translations:

The main concept is to pass an `translation` object into the editor eg.

```jsx
const translations = {
  "right_panel": {
      "edit_things_button": "Edit things",
      "move_things_button": "Move things"
  }
}

() => {

  return  (
    <Editor plugins={plugins} translations={translations} />
  )
}



```

the translations object is just my example how could it look, but in that way developer would have a full
controll over the labels all over the editor. Im not sure how the implementation of that would look, considering that
the every plugin has its own bottom drawer and other things.

## Alternative

There would be hard coded translations somewhere in the code, and developers from other coutries could submit
theirs translations. And the editor would take a `lang` prop and based on that property it would load translation for specific language.

# The most important translations/labels I'd like to be able to change are

### Right panel
- `Edit things`
- `Add things`
- `Move things`
- `Resize things`
- `Preview result`

### Left drawer
- `Add plugin to content`
- `Search plugins` - its the placeholder of search textbox
- `Content plugins`
- `No plugins found` - where we type something in search box and there aren't any plugins by that phrase.

### Slate plugin
- `Write here...`  - the text editor when its empty
- `Align left` *and other buttons title* - the titles of those buttons.

### Plugin Bottom drawer
- the titles of the button's e.g (`Toggle size`, `Content is visible` and so on).

macrozone:
i think we can use react-i18next or similar. The editor would have a new Context provider from i18next and whereever we need a label, we can use the hook from i18next.

One goal would also be to lazy load default translations (using `import()`) and we should encourage an api where users would also lazy load their custom translations. This avoids increase in bundle size for users who just view your content.

we can also provide some default translations as you suggested as well. 

The `<Editor />` would receive a new property `uiLang` or similar that would control the language of the editor ui. It shouldn't be confused with the `lang` property that controls which content-language version is displayed.

643364103:
The function is an enhancement.And it's nice!

macrozone:
devinitly still on the roadmap, personally had no time and driver for it.

rexebin:
It is a big hassle to add translations to all internal components. 

What if the `Editor` component accepts an optional function t: `(text?: string) => string`, then it gets passed to an internal context provider and default it to return given text `(text?: string) => text`. Then just prefix all the labels with the `t` function from context. No structure or prop changes to existing components. 

[...]

PR #935, opened 2021-03-26
feat(i18n, draft): add new uiTranslator prop, add new UiTranslatorContext Provider

rexebin:
## Proposed changes

As per discussion on issue #865
1. add new `uiTranslator` prop, 
2. add new `UiTranslatorContext` Provider
3. apply translation as a quick trial to the new prop.

## Types of changes

### DIFF
--- a/docs/editor.md
+++ b/docs/editor.md
@@ -90,6 +90,19 @@ That way an editing user can select the language at any time. Any cell will allw
 
 Additionaly cells can be hidden _per language_.
 
+### `cellSpacing`
+
+Takes a number or an object `{x: number, y: number}:
+
+```tsx
+cellSpacing = {
+  x: 15, // horizontal cell spacing
+  y: 20, // vertical cell spacing
+};
+```
+
+[Look at this example to see how cellSpacing affects the layout](//demo/examples/cellSpacing)
+
 ### `childConstraints` (experimental)
 
 Takes an object:
--- a/docs/editor.md
+++ b/docs/editor.md
@@ -103,6 +103,31 @@ cellSpacing = {
 
 [Look at this example to see how cellSpacing affects the layout](//demo/examples/cellSpacing)
 
+### `uiTranslator`
+
+Takes a function `(label?:string) => string | undefined`.  
+
+This prop enables i18n support. All interface labels are wrapped with this function.
+
+Below is the `uiTranslator` function provided in [the i18n demo](//demo/i18n).
+
+```typescript
+const TRANSLATIONS: { [key: string]: string } = {
+  'Edit things': '编辑',
+  'Add things': '添加',
+  'Move things': '移动',
+  'Resize things': '调整大小',
+  'Preview result': '预览模式',
+};
+
+ const uiTranslator = useCallback((label?: string) => {
+    if (TRANSLATIONS[label] !== undefined) {
+      return TRANSLATIONS[label];
+    }
+    return `${label}(to translate)`;
+  }, []);
+```
+
 ### `childConstraints` (experimental)
 
 Takes an object:
--- a/docs/editor.md
+++ b/docs/editor.md
@@ -105,7 +105,7 @@ cellSpacing = {
 
 ### `uiTranslator`
 
-Takes a function `(label?:string) => string | undefined`.  
+Takes a function `(label:string) => string`.  
 
 This prop enables i18n support. All interface labels are wrapped with this function.
 
--- a/packages/editor/src/core/Provider/index.tsx
+++ b/packages/editor/src/core/Provider/index.tsx
@@ -16,6 +16,7 @@ import { Middleware, Store } from 'redux';
 import { migrateValue } from '../migrations/migrate';
 import { updateValue } from '../actions/value';
 import { RootState } from '../types';
+import { useOptionsMemoized } from './useOptionsMemoized';
 
 type ProviderProps = {
   lang?: string;
@@ -142,29 +143,17 @@ const Provider: React.FC<ProviderProps> = ({
   }, [editorStore, lang]);
 
   // prevent options from recreating all the time
-  const optionsMemoized: Options = useMemo(() => {
-    return {
-      cellPlugins,
-      pluginsWillChange,
-      allowMoveInEditMode,
-      allowResizeInEditMode,
-      editModeResizeHandle,
-      languages,
-      childConstraints,
-      components,
-      cellSpacing,
-    };
-  }, [
-    pluginsWillChange && cellPlugins,
+  const optionsMemoized = useOptionsMemoized({
+    pluginsWillChange,
+    cellPlugins,
     allowMoveInEditMode,
     allowResizeInEditMode,
     editModeResizeHandle,
     languages,
-    JSON.stringify(childConstraints ?? {}), // its an object, we prevent unnecessary rerenders by stringify it
-    ...Object.keys(components ?? {}),
-    ...Object.values(components ?? {}), // minimize unnecessary rerenders by forcing shallow comparison of "components" object
-    JSON.stringify(cellSpacing ?? []),
-  ]);
+    childConstraints,
+    components,
+    cellSpacing,
+  });
 
   return (
     <DndProvider backend={dndBackend}>
--- a/packages/editor/src/core/Provider/index.tsx
+++ b/packages/editor/src/core/Provider/index.tsx
@@ -52,6 +52,7 @@ const Provider: React.FC<ProviderProps> = ({
   languages,
   pluginsWillChange,
   components,
+  cellSpacing,
   store: passedStore,
   middleware = [],
 }) => {
@@ -151,6 +152,7 @@ const Provider: React.FC<ProviderProps> = ({
       languages,
       childConstraints,
       components,
+      cellSpacing,
     };
   }, [
     pluginsWillChange && cellPlugins,
@@ -161,6 +163,7 @@ const Provider: React.FC<ProviderProps> = ({
     JSON.stringify(childConstraints ?? {}), // its an object, we prevent unnecessary rerenders by stringify it
     ...Object.keys(components ?? {}),
     ...Object.values(components ?? {}), // minimize unnecessary rerenders by forcing shallow comparison of "components" object
+    JSON.stringify(cellSpacing ?? []),
   ]);
 
   return (
--- a/packages/editor/src/core/Provider/index.tsx
+++ b/packages/editor/src/core/Provider/index.tsx
@@ -6,8 +6,8 @@ import BlurGate from '../components/BlurGate';
 import EditorStore, { EditorContext } from '../EditorStore';
 import { ReduxProvider } from '../reduxConnect';
 import { DisplayModes } from '../actions/display';
-import { OptionsContext } from '../components/hooks';
-import { Value, Options } from '../types/node';
+import { Value } from '../types/node';
+import { UiTranslatorContext } from '../components/hooks';
 
 import { serialzeValue } from '../migrations/serialzeValue';
 import deepEquals from '../utils/deepEquals';
@@ -15,8 +15,8 @@ import deepEquals from '../utils/deepEquals';
 import { Middleware, Store } from 'redux';
 import { migrateValue } from '../migrations/migrate';
 import { updateValue } from '../actions/value';
-import { RootState } from '../types';
-import { useOptionsMemoized } from './useOptionsMemoized';
+import { Options, RootState } from '../types';
+import OptionsProvider from './OptionsProvider';
 
 type ProviderProps = {
   lang?: string;
@@ -34,6 +34,11 @@ type ProviderProps = {
    * pass custom redux middleware:. Might get deprecated in the future
    */
   middleware?: Middleware[];
+  /**
+   * Labels are passed with this function. Use this function to replace labels for i18n support.
+   * @param label
+   */
+  uiTranslator?: (label?: string) => string | undefined;
 } & Options;
 
 const Provider: React.FC<ProviderProps> = ({
@@ -46,16 +51,11 @@ const Provider: React.FC<ProviderProps> = ({
   blurGateDisabled = false,
   blurGateDefaultMode,
   cellPlugins,
-  allowMoveInEditMode,
-  childConstraints,
-  allowResizeInEditMode,
-  editModeResizeHandle,
-  languages,
-  pluginsWillChange,
-  components,
-  cellSpacing,
   store: passedStore,
   middleware = [],
+  pluginsWillChange,
+  uiTranslator = (label) => label,
+  ...options
 }) => {
   const editorStore = useMemo(() => {
     const store = new EditorStore({
@@ -143,31 +143,33 @@ const Provider: React.FC<ProviderProps> = ({
   }, [editorStore, lang]);
 
   // prevent options from recreating all the time
-  const optionsMemoized = useOptionsMemoized({
-    pluginsWillChange,
-    cellPlugins,
-    allowMoveInEditMode,
-    allowResizeInEditMode,
-    editModeResizeHandle,
-    languages,
-    childConstraints,
-    components,
-    cellSpacing,
-  });
+
+  const uiTranslatorMemorized = useMemo(
+    () => ({
+      t: uiTranslator,
+    }),
+    [uiTranslator]
+  );
 
   return (
     <DndProvider backend={dndBackend}>
       <ReduxProvider store={editorStore.store}>
-        <OptionsContext.Provider value={optionsMemoized}>
-          <EditorContext.Provider value={editorStore}>
-            <BlurGate
-              disabled={blurGateDisabled}
-              defaultMode={blurGateDefaultMode}
-            >
-              {children}
-            </BlurGate>
-          </EditorContext.Provider>
-        </OptionsContext.Provider>
+        <UiTranslatorContext.Provider value={uiTranslatorMemorized}>
+          <OptionsProvider
+            {...options}
+            pluginsWillChange={pluginsWillChange}
+            cellPlugins={cellPlugins}
+          >
+            <EditorContext.Provider value={editorStore}>
+              <BlurGate
+                disabled={blurGateDisabled}
+                defaultMode={blurGateDefaultMode}
+              >
+                {children}
+              </BlurGate>
+            </EditorContext.Provider>
+          </OptionsProvider>
+        </UiTranslatorContext.Provider>
       </ReduxProvider>
     </DndProvider>
   );
--- a/packages/editor/src/core/Provider/index.tsx
+++ b/packages/editor/src/core/Provider/index.tsx
@@ -38,7 +38,7 @@ type ProviderProps = {
    * Labels are passed with this function. Use this function to replace labels for i18n support.
    * @param label
    */
-  uiTranslator?: (label?: string) => string | undefined;
+  uiTranslator?: (label: string) => string;
 } & Options;
 
 const Provider: React.FC<ProviderProps> = ({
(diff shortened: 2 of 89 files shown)


rexebin:
Not sure why the CI is failing. I have applied the `t` function to all labels I could find. Once settled, maybe we need to remove the existing `translations` prop in some components.

macrozone:
@rexebin thanks a lot, will review it today or tomorrow.

docs/editor.md
```diff
@@ -103,6 +103,31 @@ cellSpacing = {
...
+Below is the `uiTranslator` function provided in [the i18n demo](//demo/i18n).
+
+```typescript
+const TRANSLATIONS: { [key: string]: string } = {
+  'Edit things': '编辑',
+  'Add things': '添加',
+  'Move things': '移动',
+  'Resize things': '调整大小',
+  'Preview result': '预览模式',
+};
```
macrozone, line 121: it would be good to have a type with all the translation keys.

examples/pages/empty.tsx
```diff
@@ -15,6 +23,54 @@ const LANGUAGES: Options['languages'] = [
...
+        color={'primary'}
+        variant={'contained'}
+        style={{ marginRight: '10px' }}
+      >
+        Edit Mode
+      </Button>
+    </div>
+  );
+}
+
```
macrozone, line 73: please remove this from the "empty" example

packages/editor/src/core/Provider/index.tsx
```diff
@@ -34,6 +34,7 @@ type ProviderProps = {
    * pass custom redux middleware:. Might get deprecated in the future
    */
   middleware?: Middleware[];
+  uiTranslator: (label?: string) => string | undefined;
```
macrozone, line 37: can you add a jsdoc?

packages/editor/src/core/components/hooks/__tests__/useNodeDropLevels.test.tsx
```diff
@@ -50,7 +50,12 @@ const SAMPLE_STATE: Value = {
 describe('useNodeDropLevels', () => {
   it('tells that cells that have siblings on the right have drop level 0 there', () => {
     const Wrapper = ({ children }) => (
-      <Provider cellPlugins={[]} lang="en" value={SAMPLE_STATE}>
+      <Provider
+        cellPlugins={[]}
+        lang="en"
+        value={SAMPLE_STATE}
+        uiTranslator={(label) => label}
```
macrozone, line 57: i don't think that this test needs to be altered

packages/editor/src/ui/BottomToolbar/NodeTools.tsx
```diff
@@ -10,22 +10,22 @@ export type BottomToolbarMainBarProps = {
...
-
+    const { t } = useUiTranslator();
     return (
       <div>
         <Grid container={true} direction="row" alignItems="center">
           {icon || title ? (
             <Grid item={true}>
               <Avatar
-                children={icon || (title ? title[0] : '')}
+                children={icon || t(title ? title[0] : '')}
```
macrozone, line 20: the point here is that the avatar shows the first letter of the title if no icon is defined

so with translation it would be

`icon || t(title)[0]`

packages/editor/src/ui/DisplayModeToggle/index.tsx
```diff
@@ -42,14 +43,15 @@ export const DisplayModeToggle: React.SFC<{
...
-    { action: <ToggleEdit label={translations.edit} /> },
-    { action: <ToggleInsert label={translations.insert} /> },
-    { action: <ToggleLayout label={translations.layout} /> },
-    { action: <ToggleResize label={translations.resize} /> },
-    { action: <TogglePreview label={translations.preview} /> },
+    { action: <ToggleEdit label={t(translations.edit)} /> },
+    { action: <ToggleInsert label={t(translations.insert)} /> },
+    { action: <ToggleLayout label={t(translations.layout)} /> },
+    { action: <ToggleResize label={t(translations.resize)} /> },
+    { action: <TogglePreview label={t(translations.preview)} /> },
```
macrozone, line 54: this is one confusion that we have to discuss. Some stuff already has these translations. Should we keep them or should we migrate away?

macrozone (changes requested):
thank you very much for the work.

There are two things i am unsure about:

- what do we do with the existing translation mechanism that exists in some places?
- using the default english words as translation key is probably not a good idea. It needs some additional namespace, so that you at least know the context.

docs/editor.md
rexebin: The current structure with English words as the key is the laziest and simplest implement. Simplicity is also the main benefit of such design. It hands over 100% of responsibility to lib users and depending on requirements. The dictionary can be as partial or as complete as they want. 

Providing a complete dictionary for the lib users to translate is better, but it requires deep integrations. I have seen other libs do exactly this, most of them not bothering to provide any context descriptions, maybe because interface dictionaries are usually very small. Maybe later when there are contributions of dictionaries, react-page can switch to this solution.

As for using English words as key, I think it is only beneficial for dedicated keys when the same English words require different translations, if not happening, then separate keys posts no benefit because the English word and its translation pairs are unique. Of cause, unless I missed something.  I leave this to your good judgement.

I agree that add an extra description for describing the context could be beneficial. But with the current structure, many of them will be identical. For example, the labels for the buttons of Slate plugins, all labels go through the same render component `PluginButton`, context description needs to be added at a lower level in the plugin code. It means much deeper integrations for complete coverage.

How about changing the signature of `uiTranslator` to `(key: string, description?: string) => string`. The description will be given in English as the context in the source code as needed so that developers can catch it at their end. However, I think most of the labels are very simple and require no description.

Your existing translations object would require developers to provide it at every component as a prop, very performant and easy to use, but require a lot of work both for you and developers. If you agree with the current solution, I think removing `defaultTranslations` is the next step.

rexebin: Maybe in the label, default to English words instead of solely rely on the `t` function. So that you have the flexibility to change labels and not worry about keys being changed.

examples/pages/empty.tsx
rexebin: Will do.

packages/editor/src/core/Provider/index.tsx
rexebin: will do.

packages/editor/src/core/components/hooks/__tests__/useNodeDropLevels.test.tsx
rexebin: The test is failing because `uiTranslator` is actually a required one for the `Provider`, I will change the Provider prop signature instead then.

packages/editor/src/ui/DisplayModeToggle/index.tsx
rexebin: The existing translations object would require developers to provide it at every component as a prop, very performant and easy to use, but require a lot of work both for you and developers. If you agree with the current solution, I think removing `translations` prop is the next step.

macrozone: agreed!

packages/editor/src/core/components/hooks/__tests__/useNodeDropLevels.test.tsx
macrozone: i think it can be optional

docs/editor.md
macrozone: fully agree! let's start simple. If users need more context, let's add something like the proposed `description` prop.

packages/editor/src/ui/BottomToolbar/NodeTools.tsx
rexebin: I think to remove `t` wrapper here may be better?

rexebin:
Done most. Now just need to rebase. also separate entry for `children` PR

rexebin:
merged, rebased, checked with dev server, everything is working. 
Struggled with rebase, guess I need to use rebase more, if not proper, please help and fix it. Usually, I only merge stuff one pushed to Github.

packages/editor/src/core/types/node.ts
```diff
@@ -98,3 +98,10 @@ export type CellDrag = {
   type: 'cell';
   cell: PartialCell;
 };
+
+/**
+ * This is the ui-translator.
+ */
+export type UiTranslator = {
+  t: (label?: string) => string | undefined;
+};
```
macrozone, line 107: i would probably move it to the options.ts file instead of node.ts

packages/editor/src/core/components/Cell/ErrorCell/index.tsx
```diff
@@ -1,24 +1,27 @@
...
     <div className="react-page-cell-error">
-      <strong>An error occurred!</strong>
+      <strong>{t('An error occurred!') || 'An error occurred!'}</strong>
       <small>
         <dl>
-          <dt>Cause:</dt>
+          <dt>{t('Cause:') || 'Cause:'}</dt>
           <dd>{error.message}</dd>
-          <dt>Cell:</dt>
+          <dt>{t('Cell:') || 'Cell:'}</dt>
```
macrozone, line 18: "don't repeat yourself"

we should remove the `|| 'text'` here. instead the t-function handles the cases where the translator function does return null (or an empty string).

Remember that we control the t function that is returned from `useUiTranslator`.

packages/editor/src/core/Provider/index.tsx
```diff
@@ -137,23 +144,32 @@ const Provider: React.FC<ProviderProps> = ({
 
   // prevent options from recreating all the time
 
+  const uiTranslatorMemorized = useMemo(
+    () => ({
+      t: uiTranslator,
```
macrozone, line 149: this function here should handle the case, where `uiTranslator` returns nothing. It should return a falllback value.

macrozone (changes requested):
see comments about duplicated strings

packages/editor/src/core/types/node.ts
rexebin: Will do

packages/editor/src/core/components/Cell/ErrorCell/index.tsx
rexebin: As we discussed before, the reason I added these `|| 'text'` is to show how we can separate `key` with labels. For example, if in the future, you need to change `Cell`, keeping the key untouched but add this instead. 

However, the default `(label) => label` defeat the purpose, we actually want to change the default `t` function to return `undefined` and hard code default labels in the code, just to separate keys with labels. Keys will never be changed, but labels could be changed. 

If you want to keep it simple, then default return keys as is would be a lot simpler, with fewer flexibilities.

We don't control the `t` function, we call it and use whatever been returned. Developers control this function.

Please let me know what you think of above.

packages/editor/src/core/Provider/index.tsx
rexebin: As I mentioned in the previews comment just above, depends on what you want, `uiTranslator` could just return key or `undefined`.

packages/editor/src/core/components/Cell/ErrorCell/index.tsx
macrozone: > We don't control the t function, we call it and use whatever been returned. Developers control this function.

no, developers control the UiTranslator function.

I just really don't like to have now every string twice now in the code. It just adds noise to the code.

The t function should return a default translation if the developers UiTranslator returns null.

What about this proposel:

the `t`-functon (that geturns by the hook) has this implementation:

```
const t = (key: string) => {
 const translated = userDefinedUiTranslatorFunction(key)
 if(translated === null){
   return defaultTranslatorFunction(key) // that can be the identity to begin with and later can use some default translation object
 }
}
```

so it would always work, even if the developer forgets to translate something

rexebin:
I am totally with you. Adding default labels repeatedly is a lot is not
good. If you don’t have issue with English words as key, then my initial
implementation defaulting to return key if not found is the simplest.

rexebin:
@macrozone Please see my latest commit. 
1. I have removed all default labels `|| 'text'`, they really weren't pretty. 
2. `UiTranslator` type is moved to `options.ts` as requested.
3.  changed `uiTranslator` prop from `(labels?: string) => string | undefined` to `(label:string) => string`. So that developers have to return a string if they are using TypeScript.

To recap: 
1. `uiTranslator` prop is optional, if not provided, `Editor` component default it to `label=>label` to the label as-is. It has very little impact on the rest of the code.
2.  English labels are used as keys. This assumes English labels are unique. If they are used in multiple places in `react-page`, their translations will be the same.
3. The signature `uiTranslator` prop will force developers to always return a string hopefully. If not, I guess they will find come blank labels quickly and fix the issue themselves.

We do have full control of the default `uiTranslator`. Later, if we have a complete set of translation for a language, it is very easy to introduce a `uiLanguage` for the supported language. All we have to do is inserting dictionaries into the default `uiTranslator` function to return the correct translated labels instead of returning labels as-is. We could even accept a partial dictionary from the developer and hook it into the default `uiTranslator` function, allowing developers to override some of the translations if they wish to do so.

macrozone:
> @macrozone Please see my latest commit.
> 
> 1. I have removed all default labels `|| 'text'`, they really weren't pretty.
> 2. `UiTranslator` type is moved to `options.ts` as requested.
> 3. changed `uiTranslator` prop from `(labels?: string) => string | undefined` to `(label:string) => string`. So that developers have to return a string if they are using TypeScript.
> 
> To recap:
> 
> 1. `uiTranslator` prop is optional, if not provided, `Editor` component default it to `label=>label` to return the label as-is. It has very little impact on the rest of the code.
> 2. English labels are used as keys. This assumes English labels are unique. If they are used in multiple places, their translations will be the same.
> 3. The signature `uiTranslator` prop will force developers to always return a string hopefully. If not, I guess they will find come blank labels quickly and fix the issue themselves.
> 
> We do have full control of the default `uiTranslator`. Later, if we have a complete set of translation for a language, it is very easy to introduce a `uiLanguage` prop to switch to other supported language. All we have to do is inserting dictionaries into the default `uiTranslator` function to return the correct translated labels instead of returning labels as-is. We could even accept a partial dictionary from the developer and hook it into the default `uiTranslator` function, allowing developers to override some of the translations if they wish to do so.

can you do like i proposed here https://github.com/react-page/react-page/pull/935#discussion_r606786813 ? I think that is the best compromise

packages/editor/src/editor/Editor.tsx
```diff
@@ -55,6 +56,7 @@ const Editor: React.FC<EditorProps> = ({
   lang = languages?.[0].lang ?? 'default',
   pluginsWillChange,
   cellSpacing,
+  uiTranslator = (label) => label,
```
macrozone, line 59: as described here https://github.com/react-page/react-page/pull/935#discussion_r606786813

uiTranslator should not have a default `(label) => label`. Instead the {t} function should do that if the user provided uiTranslator does not return a value. This allows us to use default translations.

rexebin: I considered your suggestion, a lot. The t function is being provided via a context and it is a singleton. The same t function is being used everywhere, so it cannot be specific to individual use cases. I don't know how we can implement a useDefaultTranslations function to fit all.

The closest I could think of is that the t function accepts a dictionary as a second argument, and use it in place of useDefaultTranslations. However It gives more problem than it solves and it doesn't work in some cases. For example, in the PluginButton component, it renders all plugin buttons, there is so dynamic, you cannot supply a default dictionary. 

English version labels are being used as key, returning the key as-is covers the lack of translations.

macrozone: here is the t function i proposed:

```
// this is the t function returned by the useUiTranslator hook. 
const t = (key: string) => {
 const translated = userDefinedUiTranslatorFunction(key)
 if(translated === null){
   return defaultTranslatorFunction(key) // that can be the identity to begin with and later can use some default translation object
 }
}
```

`defaultTranslatorFunction would just be `(key) => key`

The problem is, that if we introduce some new label and the developers use an approach where they would not handle fallbacks, we end up with an empty label, which might affect usability negativly.

that's why i am all for handling this case on our side as it is super easy to do.

Your current solution does not do that. If the user provides a translator function that returns null for some keys, it will just not show the label at all.

rexebin: If you spin up the dev server and go to i18n demo page, you will see what I mean. I only provided translations toolbar, and I return "label+(to translate)" for everything else to show all labels are being covered.

rexebin: > here is the t function i proposed:
> 
> 
> 
> ```
> 
> // this is the t function returned by the useUiTranslator hook. 
> 
> const t = (key: string) => {
> 
>  const translated = userDefinedUiTranslatorFunction(key)
> 
>  if(translated === null){
> 
>    return defaultTranslatorFunction(key) // that can be the identity to begin with and later can use some default translation object
> 
>  }
> 
> }
> 
> ```
> 
> 
> 
> `defaultTranslatorFunction would just be `(key) => key`
> 
> 
> 
> The problem is, that if we introduce some new label and the developers use an approach where they would not handle fallbacks, we end up with an empty label, which might affect usability negativly.
> 
> 
> 
> that's why i am all for handling this case on our side as it is super easy to do.
> 
> 
> 
> Your current solution does not do that. If the user provides a translator function that returns null for some keys, it will just not show the label at all.
> 
> 
> 
> 

The signature of the uiTranslator prop require the developers to return a string. So they have to return a string, not a null or undefined. 

It is also very easy for them to always return key as-is as fallback. So they never see a blank label.

macrozone: please implement the fallback in the hook function as described above. Typescript signatures are no runtime guarantee.
 Most developers will just do something like:

```
const uiTranslator = key => MY_TRANSLATION_DICT[key]

```

Is there a show stopper from doing it like that? i described it above in multiple places. Should i push a commit with it?

rexebin: Please push a commit. I don't know how to implement your suggestions. I am keen to see how it works.

packages/editor/src/core/components/hooks/options.ts
```diff
@@ -26,6 +26,15 @@ export const OptionsContext = createContext<Options>({
...
 export const useOptions = () => useContext(OptionsContext);
 
+export const UiTranslatorContext = createContext<UiTranslator>({
+  t: (label) => label,
+});
+
+/**
+ * @returns the an object with a single `t` function for ui translations
+ */
+export const useUiTranslator = () => useContext(UiTranslatorContext);
```
macrozone, line 36: maybe this would work?

```tsx
useUiTranslator = () => {
  
  const userTranslatorFunciton =  useContext(UiTranslatorContext);
  return (key: string) => {
       return userTranslatorFunciton(key) ?? key
   }
}
````


```

macrozone: @rexebin what do you think?

packages/editor/src/editor/Editor.tsx
macrozone: showed on a comment above

rexebin: Oh, I think I understand now, are you saying if the given function returns null or undefined, the hook default to return key?

rexebin: Do you want me to have one last go or you can take care of it from here?

Sorry for the mess..

macrozone: yes! exactly, does this make sense? Sorry for being unclear and thank you for your effort!

macrozone: sure, go on, i will merge asap, rest looks good!

macrozone:
maybe i also did not show some comments of mine because of this weird review feature....

rexebin:
Done. I handled undefined and null case. Leaving empty string out, maybe in some rare cases some app wants a label to be empty.

PR #947, opened 2021-04-05
F/refactor i18n

macrozone:
@rexebin  i id some small changes. what do you think?

### DIFF
--- a/packages/editor/src/core/components/hooks/options.ts
+++ b/packages/editor/src/core/components/hooks/options.ts
@@ -2,7 +2,7 @@ import React, { createContext, useContext } from 'react';
 import EditorStore, { EditorContext } from '../../EditorStore';
 import { useSelector } from '../../reduxConnect';
 import { getLang } from '../../selector/setting';
-import { CellSpacing, Options, UiTranslator } from '../../types';
+import { CellSpacing, Options } from '../../types';
 
 import { normalizeCellSpacing } from '../../utils/getCellSpacing';
 import NoopProvider from '../Cell/NoopProvider';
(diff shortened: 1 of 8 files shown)


packages/editor/src/core/components/hooks/options.ts
```diff
@@ -26,22 +26,17 @@ export const OptionsContext = createContext<Options>({
...
+} => {
+  const { uiTranslator } = useOptions();
   return {
     t: (key: string) => {
-      const translation = userTranslatorFunction.t(key);
-      if (translation === undefined || translation === null) {
-        return key;
-      }
-      return translation;
+      return uiTranslator?.(key) ?? key;
```
macrozone, line 39: ?? does fallback to the right operand if left ist null or undefined. So the result is the same

rexebin: if the left is an empty string, it will go for the key as well?

rexebin: the `??` operator only check for null or undefined I think. So I think it is the same.

macrozone: yes, its the same. As you also correctly noticed that devs might indeed want to use empty strings in certain cases.

rexebin:
@macrozone cool refactor, I am good. Learn a few tricks here. Thank you!

rexebin:
@macrozone The `children` prop for the `Editor` component was removed from the original uiTranslator PR. It is a two-line change, do you want me to start a new PR or you would just add it during your next release? Is there a guideline for submitting adding contributor PR or you would handle it without PR? 
It was a pleasure working with you.

macrozone:
> @macrozone The `children` prop for the `Editor` component was removed from the original uiTranslator PR. It is a two-line change, do you want me to start a new PR or you would just add it during your next release? Is there a guideline for submitting adding contributor PR or you would handle it without PR?

I'll readd it and release it

> It was a pleasure working with you.

same, thank you :-)

Issue #948, opened 2021-04-06
useDebouncedCellData bug


avoskresensky:
**Describe the bug**
`useDebouncedCellData` has a bug that causes it to "forget" updates that get overrun by further updates within the 200 ms debounce timeout.

**To Reproduce**
Perform two plugin state updates within 200 ms of each other. The first one will get lost.
In order for the bug to manifest, React state batching must be in effect. This is the most common case, but to make sure, wrap those two update calls inside `ReactDOM.unstable_batchedUpdates`.

This happens because of a bug in `useDebouncedCellData`:

```
const [data, setData] = useState(cellData);
const dataRef = useRef(data);
dataRef.current = data;
```

You can do this to props, but it won't work for the state because of state batching. React won't rerender in between setState calls, so this hook won't be called, and this code won't have a chance to update the state ref. The ref update must happen right before setState inside the onChange function that gets returned to the user:

```
const [data, setData] = useState(cellData);
const dataRef = useRef(data);

const onChange = useCallback((update) => {
    dataRef.current = {...dataRef.current, ...update};
    setState(dataRef.current);
}, [setData]);

return [data, onChange];
```

It makes a lot of sense to return `dataRef` instead of just `data` from this hook, but that's a different story.

Edit:

Another, somewhat smaller problem with useDebouncedCellData is that it returns a referentially unstable onChange function which is not what you usually expect from a hook. onChange needlessly depends on `data` which can be easily helped since there is a ref there.

macrozone:
can you fix it?

the current code is the result of dealing with these kind of issues, and it worked very well so far, but is probably far from beeing correct..

avoskresensky:
> can you fix it?

Sure. Will send a PR shortly.

PR #949, opened 2021-04-07
bug(hook): useDebouncedCellData fix

avoskresensky:
Fixed useDebouncedCellData as discussed in #948.

Added 2 unit tests to verify the fixes of two outlined problems + another one ("outside changes") to make sure the original corner case logic works as before.

## Types of changes

## Checklist

## Closing issues

#948

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7133,6 +7133,13 @@ jest-pnp-resolver@^1.2.2:
   resolved "https://registry.yarnpkg.com/jest-pnp-resolver/-/jest-pnp-resolver-1.2.2.tgz#b704ac0ae028a89108a4d040b3f919dfddc8e33c"
   integrity sha512-olV41bKSMm8BdnuMsewT4jqlZ8+3TCARAXjZGT9jcoSnrfUnRCqnMoF9XEeoWjbzObpqF9dRhHQj0Xb9QdF6/w==
 
+jest-react-hooks-shallow@^1.5.1:
+  version "1.5.1"
+  resolved "https://registry.yarnpkg.com/jest-react-hooks-shallow/-/jest-react-hooks-shallow-1.5.1.tgz#d522381061bb9dff5671f84b053d0b3f1cbe37f9"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -10396,6 +10403,15 @@ react-transition-group@^4.4.0:
...
 
+react@^16.8.0:
+  version "16.14.0"
+  resolved "https://registry.yarnpkg.com/react/-/react-16.14.0.tgz#94d776ddd0aaa37da3eda8fc5b6b18a4c9a3114d"
+  integrity sha512-0X2CImDkJGApiAlcf0ODKIneSwBPhqJawOa5wCtKbu7ZECrmS26NvtSILynQ66cgkT/RJ4LidJOc3bUESwmU8g==
+  dependencies:
+    loose-envify "^1.1.0"
+    object-assign "^4.1.1"
+    prop-types "^15.6.2"
+
```
macrozone, line 10414: this is a dep of jest-react-hooks-shallow, maybe that is outdated?

it probably has no side effect though

macrozone:
@avoskresensky thank you so much for taking the time to tackle this and add a proper test!

i'll check locally and merge into beta. Goal is to release beta soon (next week or so)

macrozone:
there is one problem that  undo might be broken now:

https://user-images.githubusercontent.com/1972353/114039642-373fd500-9883-11eb-834d-7be8335fea00.mov


notice how the undo button goes back to be "at the end" (so no redo is possible).

reason is that after undo, it seems to send out another onChange event that will erase the future

yarn.lock
avoskresensky: Those three were already a part of the node_modules. Their versions remained intact.

avoskresensky:
@macrozone , that was a problem inside the slate plugin: onChangeDebounced callback was missing a dependency (`data.slate`) and now that we made `props.onChange` referentially stable, it stopped updating the closure and notUndoable flag was set incorrectly.

I updated the dependency which fixed the immediate problem, but the slate plugin obviously needs some more work:

- The debouncing logic can probably go now that we have it "upstairs"
- There is another example of the dreaded "stale state" anti-pattern:

```
const [value, setValue] = useState<Node[]>(data?.slate);
const valueRef = useRef(value);
valueRef.current = value;
```

macrozone:
works good as far as i can tell! thank you very much @avoskresensky

Issue #945, opened 2021-04-05
Trouble focusing the Background plugin


avoskresensky:
**Describe the bug**
Certain layouts make focusing the underlying background plugin unintuitive.

**To Reproduce**
1. Add a background plugin.
2. Add 4 plugins to it: two columns with two plugins each, one under another.
3. Make sure the columns are of uneven height
4. The blank space under the shortest column (within the background plugin) is not clickable. Clicks won't select the background plugin.

![image](https://user-images.githubusercontent.com/10179521/113563459-961ff880-95bc-11eb-9a7c-247e1f7ccc05.png)

Another strange (and maybe related) thing about this layout is that if you leave only one plugin in a column, it will consume the entire height of the column.

![image](https://user-images.githubusercontent.com/10179521/113563611-dda68480-95bc-11eb-83d0-81cc0a0e7db5.png)

macrozone:
I know this problem, so i introduced these "tabs":

<img width="934" alt="Bildschirmfoto 2021-04-05 um 12 22 19" src="https://user-images.githubusercontent.com/1972353/113564375-94c8ed00-9609-11eb-9f88-17fe2d9c38ef.png">

problem is, that if you specify "overflow: hidden" on your background plugin, you won't see it.


The stretching of the cells to fill the parent is intentional, without it dragging and dropping cells next to these cells might be awkward, at least i think i did it because of that (should have documented it as an issue though).

Having said that, i also don't fully like the current solution, so the idea i have is to show the full hierarchy of the selected cell as tabs in the bottom toolbar (or in a similar way) and you can select any parent directly.

So something like:

`[ **Image** ] [ Background ]`


(or any other parents if its deeply nested)

Additionaly we could allow to select rows (or non-plugin cells) as well. This way we could also move / duplicate / delete rows. This might be handy if we have configurable rows.

Also we could enable to select "non-plugin" cells as well (the ones that lead to this issue: https://github.com/react-page/react-page/issues/943)

So the full hierarchy on the left is actually (if you would select one of the images):


`[ **Image** ][Row][Cell][Row][ Background ]`

avoskresensky:
@macrozone, I have two problems with this artificial "stretching" at this point. Firstly, it's inconsistent: on the second screenshot it's present, but on the first one it doesn't really stretch -- it just prevents the clicks from selecting the containing layout plugin I guess.

[...]

PR #950, opened 2021-04-07
bug(ui): trouble focusing the Background plugin

avoskresensky:
Allow clicks on layout plugin body to bubble up and be caught by the focusing logic.
Prevent blurAll when clicking on resize handle or when done resizing.

## Proposed changes

Stop propagation in their respective elements.
Handle phantom onClick for "react-page-row" separately -- that event is being fired when resizing stops.
Remove blurAll from the row (it was hit by the phantom onClick).

## Types of changes

## Checklist

## Closing issues

#945

### DIFF
--- a/packages/editor/src/core/components/Row/index.tsx
+++ b/packages/editor/src/core/components/Row/index.tsx
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import React from 'react';
 import { useMeasure } from 'react-use';
 import { isRow, Node, Row } from '../../types/node';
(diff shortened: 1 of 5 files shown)


packages/editor/src/core/components/Row/index.tsx
```diff
@@ -57,8 +56,7 @@ const Row: React.FC<{ nodeId: string }> = ({ nodeId }) => {
             rowHasInlineChildrenPosition
           ),
         })}
-        style={{ position: 'relative', borderColor: 'red' }}
```
macrozone, line 60: hehe, thx for removing that

Issue #962, opened 2021-04-15
Performance problems of the initial rendering of various editor modes


avoskresensky:
**Describe the bug**
The performance of the initial rendering of various editor modes seems to significantly differ.
By initial rendering I mean the moment a user switches between modes using the toolbar.

According to our observations and some measurements, the winner, interestingly enough, is the "edit" mode, while the absolute loser is, interestingly enough, the "resize" mode.

We clearly notice the difference with some huge pages we have here in our internal app, which has a bunch of custom plugins. So to rule out the plugin interference, I'll try to demonstrate using a standard example page.

The measurements show that the difference in performance is not caused by how long it takes for the component tree to reconcile (render), but how long it takes to commit it, which seems to be a function of the number of state altering side effects.

The best way to demonstrate the problem is by using the Timings feature of the Performance tab of Chrome Dev Tools. It uses the Timing API which was removed from the latest versions of React, so in order to reproduce the attached measurements you'll need to downgrade your React to 16.x (I used 16.14).

In the below screenshots I'm using cellSpacing example and simply switching between editor modes.
It makes a difference which mode you are switching *from*, so I took all measurements switching from "preview" to the target mode. When measuring "preview" performance I was switching from "edit".

I have some Chrome plugins enabled, so absolute numbers are useless. However, when compared to each other they seem to make sense.
The reconciliation phase is about the same in all the measurements, but the commit phase differs greatly.

The "edit" mode is the only one that doesn't seem to have any state altering side effects. Here is the graph for it:

![image](https://user-images.githubusercontent.com/10179521/114810135-50072980-9d60-11eb-825d-c6ca3c951ed1.png)

"Insert" and "preview" modes seem to be about 50% slower, most likely because of some state altering side effect (cascading update warning):

![image](https://user-images.githubusercontent.com/10179521/114817144-54d2da00-9d6e-11eb-8a36-ed40519a0c6a.png)

The "move" mode is usually about 2 times slower, again with the cascading update warning:

![image](https://user-images.githubusercontent.com/10179521/114810671-6cf02c80-9d61-11eb-8c62-42fa2a74d2d1.png)

And here's our champ, the "resize" mode:

![image](https://user-images.githubusercontent.com/10179521/114815801-d07f5780-9d6b-11eb-9e8b-43ae50c4e0d0.png)

It's usually 10-15 times slower than "edit"!
It shows the same mammoth number regardless of whether you enter or exit it.
[...]

PR #968, opened 2021-04-16
fix: switching between resize mode and other modes is slow

macrozone:
fixes https://github.com/react-page/react-page/issues/962

## Types of changes

## Checklist

- [] I have added necessary documentation (if appropriate)

## Closing issues

closes https://github.com/react-page/react-page/issues/962


avoskresensky:
@macrozone, just noticed that background plugin dragging in layout mode was affected by this change.

Now the background plugin doesn't show the move cursor in layout mode when hovering over its body and it doesn't drag.

I think it has to do with "react-page-cell-draggable" not being added after this modification.

PR #971, opened 2021-04-20
fix/962: restore background dragging + dragging by handle styles

### LINKED ISSUE #962: Performance problems of the initial rendering of various editor modes

avoskresensky:
## Proposed changes

Fixes a problem with background plugins not responding to dragging in layout mode.
Also adds styles for highlighting a cell that is dragged by the cell handle.

## Types of changes

## Checklist

## Closing issues

Closes #962

### DIFF
--- a/packages/editor/src/core/components/Cell/Draggable/index.css
+++ b/packages/editor/src/core/components/Cell/Draggable/index.css
@@ -110,7 +110,9 @@
       z-index: 4;
     }
 
-    .react-page-cell-draggable-is-dragging {
+    .react-page-cell-draggable-is-dragging,
+    .react-page-cell-handle-is-dragging + div > .react-page-cell-droppable > .react-page-cell-draggable,
+    .react-page-cell-handle-is-dragging + div > .react-page-cell-droppable > .react-page-cell-draggable-in-edit {
       opacity: 0.4;
       outline: none;
     }
--- a/packages/editor/src/core/components/Cell/Draggable/index.tsx
+++ b/packages/editor/src/core/components/Cell/Draggable/index.tsx
@@ -34,20 +34,20 @@ const Draggable: React.FC<Props> = ({ isLeaf, children, nodeId }) => {
     <>
       {previewElement}
       <div
+        ref={isLayoutMode ? dragRef : undefined}
         style={{
           height: '100%',
         }}
         className={classNames({
-          'react-page-cell-draggable-in-edit': options.allowMoveInEditMode,
-          'react-page-cell-draggable':
-            isLayoutMode && !options.allowMoveInEditMode,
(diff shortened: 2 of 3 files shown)


macrozone:
👍

PR #1002, opened 2021-05-17
Fix: Uniform versions

PeterKottas:
A not yet working fix for the uniforms versioning.

### DIFF
--- a/packages/editor/yarn.lock
+++ b/packages/editor/yarn.lock
@@ -299,7 +299,7 @@ object-assign@^4.1.1:
   resolved "https://registry.yarnpkg.com/object-assign/-/object-assign-4.1.1.tgz#2109adc7965887cfc05cbbd442cac8bfbb360863"
   integrity sha1-IQmtx5ZYh8/AXLvUQsrIv7s2CGM=
 
-prop-types@^15.5.10, prop-types@^15.5.8, prop-types@^15.6.0, prop-types@^15.7.2:
+prop-types@^15.5.10, prop-types@^15.5.7, prop-types@^15.5.8, prop-types@^15.6.0, prop-types@^15.7.2:
   version "15.7.2"
   resolved "https://registry.yarnpkg.com/prop-types/-/prop-types-15.7.2.tgz#52c41e75b8c87e72b9d9360e0206b99dcbffa6c5"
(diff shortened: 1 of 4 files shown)


packages/editor/yarn.lock
```diff
@@ -541,49 +541,37 @@ tslib@^2.1.0:
...
+    uniforms "^2.6.9"
 
-uniforms@3.0.x:
-  version "3.0.1"
-  resolved "https://registry.yarnpkg.com/uniforms/-/uniforms-3.0.1.tgz#501516bec1e556d815275e4bf8ea5b18b1f38078"
-  integrity sha512-ULYBRs0GUc1HGKY+TpOIeYQnCFdy25CDUFkPqrlYlNU45YnommuadFtl3o6TyeOytY4z7M6Hy0SgDSeWYYZoLQ==
+"uniforms@^2.6.0 <=3.4.0", uniforms@^2.6.9:
+  version "2.6.9"
+  resolved "https://registry.yarnpkg.com/uniforms/-/uniforms-2.6.9.tgz#0741532cf82331d5e432f0fb5547ac24713965ba"
+  integrity sha512-RnKx276c4xS5yoo8MxH2EtZbmzpsb5vFWP+VYCMC8tH/VSK4jlVBgCbCTADbMlcojwnAOGjUrA+m3s4tacdLbg==
```
macrozone, line 568: it seems it downgraded the version to 2.6.9. that is probablyy the problem you are facing. Altough i am not sure if this yarn.lock file is used at all (maybe it is used in the demo project, as it is a lerna mono-repo). 

nevertheless, you can safely revert this yarn.lock change. the change in the package.json is the only required change.

optionally we can add resolutions field to force a version for the demo project. npm installations are not affected by this. they only look for the version in the package.json (and their own local lock files of couse)

PeterKottas: Yeah, I was thinking the same. But wondering why it did the downgrade with a fresh install in the first place. I will probably read up on this. Because if this is the case, then people who just clone and install won't be able to work on this. Also, that means the package is not in fact compatible with 2.6.9 (or to be more precise the types aren't ) so I am wondering if we should not abandon the 2nd major version support. Would this even be a problem for people? It seems reasonable. 

I don't like the resolutions approach, it often backfires when people forget it's there.

macrozone:
fixed in master

Issue #1012, opened 2021-05-31
inserting new cells no longer scrolls to the cell


macrozone:
**Describe the bug**

clicking on insert mode(+) then clicking on a plugin adds a new cell to the end, but does no longer scorll down

**To Reproduce**
see above

**Expected behavior**

it should scroll down

PR #1013, opened 2021-05-31
fix: insert new does not focus cell

macrozone:
fixes https://github.com/react-page/react-page/issues/1012

@PeterKottas that uses your new focusAfter option


PeterKottas:
Looks good. I am not using the drawer, so this slipped through the cracks. Now (with sensible default value) I quite like that you can control the behavior.

Issue #982, opened 2021-04-30
handle the slash '/' command to open the insert plugin sidebar 


karthick3018:
**Is your feature request related to a problem? Please describe.**
Since all the editor related options like notion , rich editor , slab handles the slash command people tends to go for "/" command  
**Describe the solution you'd like**
We can have the global listener and handle the slash command

macrozone:
nice idea.

Its probably good to make it customizable. E.g. on swiss-german keyboards like mine, there is no dedicated `/`-key

karthick3018:
@macrozone unlike slash we can have a control key and any letter combination ,I would love to work on the sutff if assigned

macrozone:
@karthick3018 sure thing! PRs are highly welcomed!

the slate (rich text) plugin has hotkey support, maybe you can use that as inspiration: https://github.com/react-page/react-page/blob/master/packages/plugins/content/slate/src/components/hotkeyHooks.ts

karthick3018:
sure 🎉

macrozone:
--> this currently leads to the situation that you can no longer type in a "/" char in the slate (rich text) editor

karthick3018:
@macrozone  have you fixed this issue ? I could reproduce the issue in current live version !

macrozone:
its fixed in live, or what do you mean?

we no longer open the insert plugin sidebar when any cell is focused

karthick3018:
Imagine like while being active in the rich editor I need to insert a divider plugin so I can type "/" to get the cell plugins . "we no longer open the insert plugin sidebar when any cell is focused" this one also make sense it just depends on the scenario of our application .

macrozone:
@karthick3018 I also thought about that, so you can now press ESC to unselect any cell, which makes it a bit easier to open the toolbar

karthick3018:
yeah @macrozone  current version looks so cool 🎉🎉

PR #1020, opened 2021-06-04
handle the slash '/' command to open the insert plugin sidebar #982

karthick3018:
To Open the Cell Plugins on pressing the slash "/" key
 
## Types of changes

## Checklist

### DIFF
--- a/packages/editor/src/core/components/Cell/InsertNew.tsx
+++ b/packages/editor/src/core/components/Cell/InsertNew.tsx
@@ -31,6 +31,25 @@ const InsertNew: React.FC<{ parentCellId?: string }> = ({ parentCellId }) => {
     },
   });
 
+  React.useEffect(() => {
+    window?.addEventListener('keydown', handleKeyDown);
+    return () => {
+      window?.removeEventListener('keydown', handleKeyDown);
+    };
+  }, []);
+
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (e?.key === '/') {
+      handleInsertMode(e);
+    }
+  };
+
+  const handleInsertMode = (e) => {
+    e.stopPropagation();
+    setReferenceNodeId(parentCellId);
+    setInsertMode();
+  };
+
   if (isPreviewMode) return null;
   return (
     <div
@@ -44,11 +63,7 @@ const InsertNew: React.FC<{ parentCellId?: string }> = ({ parentCellId }) => {
         minWidth: 120,
         margin: 'auto',
       }}
-      onClick={(e) => {
-        e.stopPropagation();
-        setReferenceNodeId(parentCellId);
-        setInsertMode();
-      }}
+      onClick={handleInsertMode}
     />
   );
 };
--- a/packages/editor/src/core/components/Cell/InsertNew.tsx
+++ b/packages/editor/src/core/components/Cell/InsertNew.tsx
@@ -31,25 +31,6 @@ const InsertNew: React.FC<{ parentCellId?: string }> = ({ parentCellId }) => {
     },
   });
 
-  React.useEffect(() => {
-    window?.addEventListener('keydown', handleKeyDown);
-    return () => {
-      window?.removeEventListener('keydown', handleKeyDown);
-    };
-  }, []);
-
-  const handleKeyDown = (e: KeyboardEvent) => {
-    if (e?.key === '/') {
-      handleInsertMode(e);
-    }
-  };
-
-  const handleInsertMode = (e) => {
-    e.stopPropagation();
-    setReferenceNodeId(parentCellId);
-    setInsertMode();
-  };
-
   if (isPreviewMode) return null;
   return (
     <div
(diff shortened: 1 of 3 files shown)


karthick3018:
fixes #982

packages/editor/src/core/components/Cell/InsertNew.tsx
```diff
@@ -31,6 +31,25 @@ const InsertNew: React.FC<{ parentCellId?: string }> = ({ parentCellId }) => {
     },
   });
 
+  React.useEffect(() => {
+    window?.addEventListener('keydown', handleKeyDown);
+    return () => {
+      window?.removeEventListener('keydown', handleKeyDown);
+    };
+  }, []);
```
macrozone, line 39: handleKeyDown is used before define, you should move the declaration up. 
You can declare it in `useEffect` directly. Make sure to add `handleInsertMode` to the deps array then.

packages/editor/src/core/components/Cell/InsertNew.tsx
```diff
@@ -31,6 +31,25 @@ const InsertNew: React.FC<{ parentCellId?: string }> = ({ parentCellId }) => {
...
+
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (e?.key === '/') {
```
macrozone, line 42: can we make this configurable? (could also be done later)

macrozone (changes requested):
The effect was added in `InsertNew`. This component is mounted multiple times (e.g. in cell plugins that take other cells). Additionaly this component has a reference node id (of the parent cell), which is responsible to show the plugins that can be added to that parent cell.

The key listener effect should therefore be moved in a component that is only mounted once

packages/editor/src/core/components/Cell/InsertNew.tsx
karthick3018: can you specify which file ? for better understanding

packages/editor/src/core/components/Cell/InsertNew.tsx
karthick3018: I didn't get this point throw an example like how you're expecting

macrozone: i am open for a suggestion, but we can omit it for the time being

packages/editor/src/core/components/Cell/InsertNew.tsx
macrozone: i would move it somewhere in EditableEditor

karthick3018: packages/editor/src/editor/EditableEditor.tsx -> this file doesn't have the redux connect to handle the useSetInsertMode hook . packages/editor/src/core/components/Editable/index.tsx -> shall we move the logic to this file?

macrozone: best wrap it in a component (`<GlobalHotkeys />`) and add that to the `<EditableEditor />` inside the `<Provider />`

then this `GlobalHotkeys` just returns null, but has a `useEffect` hook.

karthick3018: @macrozone updated !

packages/editor/src/core/components/Cell/InsertNew.tsx
```diff
@@ -31,6 +31,25 @@ const InsertNew: React.FC<{ parentCellId?: string }> = ({ parentCellId }) => {
...
+      handleInsertMode(e);
+    }
+  };
+
+  const handleInsertMode = (e) => {
+    e.stopPropagation();
+    setReferenceNodeId(parentCellId);
+    setInsertMode();
+  };
+
```
macrozone, line 52: you can now undo all these changes here

macrozone (changes requested):

packages/editor/src/core/components/Cell/InsertNew.tsx
karthick3018: @macrozone all changes done !!

karthick3018:
thanks @macrozone

macrozone:
@karthick3018 sorry, the feature was not release as the linting failed. Sorry i did not check that and fix that on master.

I will probably need to introduce precommit hooks to mitigate that as sometimes devs have no linter running locally

karthick3018:
@macrozone  do I need to fix anything and commit to solve those linting ?

macrozone:
> @macrozone do I need to fix anything and commit to solve those linting ?

no, just did it. (i just ran `yarn lint --fix`)

karthick3018:
@macrozone relaxing comment!!! thanks

macrozone:
@karthick3018  ;-)

travis has some problems currently as it seems, might take some more minutes to build (usually it builds in under 5min)

karthick3018:
@macrozone hope it will resolve soon , event listener PR shouldn't cause this much trouble !!lolz

PR #1178, opened 2022-04-28
feat(scope): Add alternative option to image plugin

kienmatu:
## Types of changes

## Checklist

- [ ] Lint and unit tests pass locally with my changes

### DIFF
--- a/packages/plugins/content/image/src/createPlugin.tsx
+++ b/packages/plugins/content/image/src/createPlugin.tsx
@@ -19,7 +19,7 @@ const createPlugin = (settings?: ImageSettings): CellPlugin<ImageState> => {
       ),
     },
     Renderer: mergedSettings.Renderer,
-    id: 'kienmatu/ory/editor/core/content/image',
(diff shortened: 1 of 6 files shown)


packages/plugins/content/image/src/createPlugin.tsx
```diff
@@ -19,7 +19,7 @@ const createPlugin = (settings?: ImageSettings): CellPlugin<ImageState> => {
       ),
     },
     Renderer: mergedSettings.Renderer,
-    id: 'ory/editor/core/content/image',
+    id: 'kienmatu/ory/editor/core/content/image',
```
macrozone, line 22: careful, this needs to be undone

macrozone (changes requested):

macrozone:
thx for your contributions! one minor thing to fix

packages/plugins/content/image/src/createPlugin.tsx
kienmatu: Okay, fixed!

PR #1219, opened 2022-07-31
React 18 upgrade

gerom76:
Author: Gerom76 <gerom76@gmail.com>

    React 18 upgrade

## Proposed changes
    Workaround for Slate editor deadlock while entering a text.
    Just consulting changes for the final solution.

## Types of changes

Apart from refactoring code for React 18 conventions,
in order to mitigate Slate editor deadlock I had to:
- disable Lazy loading for SlateProvider 
- disable hook usePluginIsDisabled functionality

I'd like to consult on how to recover such features again.


macrozone:
very nice, thx @gerom76 !

some notes:

- we should still allow react 17 at least, so peer deps should be lowered, i guess the code should work also with react 17
- we need to re-enable lazy-loading of slateprovider, as it is very big
- not sure why usePluginIsDisabled, but i guess it is used by some projects

gerom76:
> very nice, thx @gerom76 !
> 
> some notes:
> 
> * we should still allow react 17 at least, so peer deps should be lowered, i guess the code should work also with react 17
> * we need to re-enable lazy-loading of slateprovider, as it is very big
> * not sure why usePluginIsDisabled, but i guess it is used by some projects

Well, I have reverted the peer deps to the previous state. Regarding the slateprovider and hook, the purpose of this MR is to collaborate in developing a solution for the problem pinpointed right now. Could you run the code and provide some hints on how to restore disabled temporary features?

gerom76:
> * we need to re-enable lazy-loading of slateprovider, as it is very big

Hello @macrozone
Have you tried running my branch? Have you got any idea how to mitigate deadlock in the Slate plugin?
I analyzed the bundled package. The slate editor is around  88 kB, which is not a big problem for deployed apps. Did I miss something regarding the Slate plugin size? 

image

macrozone:
hey thanks again for your contribution, i did not have the time yet to look into it, but i will do it next week! thx for patience!

CapsE:
Any news on this? I could really use react-page in my react 18 project. Right now I'm using a seperate project/build job just for the page editor but that makes sharing components harder than it needs to be...

CapsE:
@macrozone Sorry for spamming but I tried to convert my app into a progressive web app and struggled for hours (without success) to get my external react-page application to work within the same PWA because of the workaround I had to build.

Can you give a realistic timeline on when we'll be able to use react-page with react 18? If it's still months in the future I'd be forced to switch to a different library which would be a pity because I really like what you've done so far.

macrozone:
hey @gerom76 i will continue on this today and hopefully finish it

macrozone:
thx, i will take it from here!

macrozone:
@gerom76 , @CapsE react-page now works with react18  in latest version! (5.1.4)

CapsE:
That's awesome news! Thanks for your hard work! :)

PR #1268, opened 2023-01-26
fix: language switch debounce bug

jaymaycry:
Fix language switch debounce bug

## Proposed changes

Currently when switching language the first change you make in the second language affects the contents of the first language because of currentLanguage not changing due to debounce issues. As a quick fix I introduce a ref that hols the lang and behaves correctly then.

## Types of changes

## Checklist

- [ ] Lint and unit tests pass locally with my changes


macrozone:
better fix coming

